PR #148, opened 2014-10-16
Frontend apache24

vkuznet:


geneguvo:
Hi,

This looks good, but I'll not pick the changes for converting DNs to the new format. The conversion is not simple like that. You'd need to convert non-ascii chars to UTF8 and escape special characters with backslash, at least. At the top of my head, I see the procedure you wrote would not convert service DNs like "/DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch" properly.

I'd not try writing our own routine to convert DNs from the legacy format to RFC 2253. Rather, let's pick it from elsewhere we can trust. There's a lot of bits everywhere on the grid and on CMS that do DN matching, so let's keep that migration separate to the future.

I'm not picking the couchdb change either since we're not making any changes to the couch setup yet.

Thanks, Diego.

vkuznet:
Diego,
I'm fine with not including DN changes. I did that because I saw the changes in
default DN parsing, but I am ok to use any other proper solution (if we need to)
and perform migration at a different time (if any).
Best,
Valentin.

PR #156, opened 2015-01-14
Add new reqmgr2 front page and acdc cleanup

ticoann:
Please review. This is for HG1502


geneguvo:
Hi,
I'm a bit confused by the 24 x 4 hours bits:
60 * 60 * 24 # every 4 hours

Code says 24, but comment tell 4 hours. I'm assuming the right one is 4 hours, at least that's what we are using for the long cycle in production right now, so I'd expect making the cleaning at that interval too.

In any case, please confirm what is the desired interval for the short/long cycles and for the cleaning.

Cheers, Diego.

PR #162, opened 2015-02-11
couch 1.6 update

ticoann:
Need for couch 1.6 update

### DIFF
--- a/wmagent/local.ini
+++ b/wmagent/local.ini
@@ -41,3 +41,16 @@ file = deploy_project_root/couchdb/logs/couch.log
 ;cert_file = deploy_project_root/couchdb/certs/cert.pem
 ;key_file = deploy_project_root/couchdb/certs/key.pem
 ;cacert_file = deploy_project_root/couchdb/certs/cert.pem
+ssl_certificate_max_depth = 10
+
+[replicator]
+;cert_file = deploy_project_root/couchdb/certs/cert.pem
+;key_file = deploy_project_root/couchdb/certs/key.pem
+;cacert_file = deploy_project_root/couchdb/certs/cert.pem
+ssl_certificate_max_depth = 10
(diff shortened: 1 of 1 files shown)


ticoann:
@vkuznet, please review

vkuznet:
Please add ssl_certificate_max_depth = 3 at the end of replicator section. This is required to handle enclosed certificates, like CERN proxies.

ticoann:
Thanks Valentin, I missed that. I updated the patch. Also it seem following are useful parameters to set as well. Should we set them?
use_checkpoints = true
checkpoint_interval = 5000
Here is the reference
https://people.apache.org/~dch/snapshots/couchdb/20140715/new-docs-build/html/config/replicator.html

vkuznet:
Seangchan,
yes they're good to include, but I'll make a comment above them saying that
they're introduced with CDB1.6
Valentin.

geneguvo:
Hi guys,
I'll not include this PR while we haven't concluded the replication tests with CDB16. I know, for instance, you'd need:
use_checkpoints = false
max_replication_retry_count = infinity

We may want to fine tune it too, including things like:
socket_options = [{keepalive, true}, {nodelay, true}]
and increasing size of send/recv buffers, etc..

Besides, "use_checkpoints" was introduced in CDB16. So we should not pick it before the wmagents are ready to move.

Finally, although "ssl_certificate_max_depth = 3" would work for the current cmsweb chain of certificates, this might eventually change. Default reasonable value for max depth is usually 10.

Cheers,
Diego.

ticoann:
Thanks Diego, I updated the patch as you suggested. We will run the test first, so please don't merge yet.

amaltaro:
Diego, we have been running all ours tests with this patch and it's good to be merged now.

PR #166, opened 2015-02-19
CMS data popularity app and Victor the cleaner

rcaspart:
Dear all,

This pullrequest includes the configuration files and deploy and managment scripts for the CMS data popularity services 'Victor the cleaner' and the frontend for the data popularity databases.

For these services the ports 9107 and 9106 were chosen.

Feel free to comment or make any suggestions to this.

Cheers,
Rene


rcaspart:
Hi all,

as suggested by @nikmagini (in the pull request for the cmsdist repository) I appended "-web" to the names of both services.

Cheers,
Rene

geneguvo:
Hi,
We've reviewed all this and it is very good, but as usual we've made a few changes/adjustements we'll report later.

We've now reached the point where we need the databases in place so we can start/run the services. We need you guys to send me the DB endpoints for dev, testbed and prod with their corresponding secrets following the instructions I've sent be e-mail.

We also need to know the requirements for  these services in order to define where to run them. On the meeting on the O&C week last year you mentioned but I cannot remember the numbers anymore :-(. Some estimation on the number of requests per day and the amount of used memory would probably be already enough.

Thanks,
Diego / Bruno.

geneguvo:
Hi,
We've noticed the following configuration on settings.py of the services:
SECRET_KEY = '*********'
(the stars "*" in place of the key to avoid exposure)

What is this secret about? If it needs confidentiality, we have to store it under the secrets area and should by no means make it public on github! (everything on github is read public unless you pay for it)

rcaspart:
Hi Diego,

This configuration is used by Django for each Django project. As I understand it from the documentation [*], it is used as a seed for secret-key hashing algorithms.
I am not entirely sure if there is any (and if there is any how large the impact is) impact on the security of the services by having this setting in a public place.

Cheers,
Rene

[*] https://docs.djangoproject.com/en/1.4/ref/settings/#std:setting-SECRET_KEY

giffels:
Maybe @DomenicoGiordano can comment or we simply store it under the secrets area.

Cheers,
Manuel

DomenicoGiordano:
Hi,

reading the documentation[*] I think you can leave it empty if you use "django-admin.py startproject "
because it creates it for you. 
Otherwise, have it fixed (as in your case) implies that the random sequence is fixed forever (and can be regenerated).
Another option is that you fill this field at every restart of the service, putting a random string in it, as also suggested in [*]

[*]
SECRET_KEY¶

Default: '' (Empty string)

A secret key for this particular Django installation. Used to provide a seed in secret-key hashing algorithms. Set this to a random string – the longer, the better. django-admin.py startproject creates one automatically.

nikmagini:
Hi,

we agreed with Diego to generate random keys on each backend, it shouldn't affect anything since we're not using sessions/cookies in Django for popdb and victor; to be verified once it's on testbed.

Cheers
N.

nikmagini:
Hi,

additional comment: I noticed that in fact popdbweb/victorweb set some cookies in the browser:
__utma __utmb etc...
From a rapid search, they are from google analytics rather than django however.

Cheers
N

PR #171, opened 2015-03-11
add reqmgr2 config in workqueue and wmagent

ticoann:


geneguvo:
Hi Seangchan,

Is this PR something I'm supposed to pick for HG1404? I know you have not requested it in https://svnweb.cern.ch/trac/CMSDMWM/ticket/4320, but it is worth double checking...

If we'd pick it, then this is basically switching workqueue from using ReqMgr2 instead of ReqMgr1, right? (see https://github.com/dmwm/deployment/pull/171/files#diff-a7cae201819f3cb4f9477308694a7aa1R44). This shall be a big change... please confirm that's what you intend here, because from the title I assume you only intend to add the extra config, not yet to switch.

On the review of the PR itself, please avoid hardcoding hostnames on the config file. Instead, we shall use the dev/preprod/prod variants. See for instance the usage of the "@@COUCH_HOST@@" (https://github.com/dmwm/deployment/blob/master/reqmgr2/deploy#L40). If there are too many differences in the config from dev/preprod/prod, we should keep each configuration on their own separate files config-{dev,preprod,prod}.py, like workqueue is doing at the moment, for instance.

Cheers,
Diego.

geneguvo:
Let me do the change to remove the hardcoded machine names myself.

In any case, please let me know if you want it to be included for HG1504 or not, and that the switch to reqmgr2 was on purpose or not.

ticoann:
Sorry, Diego. I wasn't clear about that. Yes, this intend to go HG1504 change. If you haven't remove the hard coded part, I will do that. Thank you very much for all the help.

geneguvo:
And how about the switch to reqmgr2? is it intentional ?

ticoann:
> If we'd pick it, then this is basically switching workqueue from using ReqMgr2 instead of ReqMgr1, right? (see https://github.com/dmwm/deployment/pull/171/files#diff-a7cae201819f3cb4f9477308694a7aa1R44). This shall be a big change... please confirm that's what you intend here, because from the title I assume you only intend to add the extra config, not yet to switch.

It is not the complete switch. But READ operation will switch to ReqMgr2. WRITE operation has flag to switch but I am not set the flag to True yet.
https://github.com/ticoann/deployment/blob/reqmgr2_transition/workqueue/GlobalWorkQueueConfig-dev.py#L49

ticoann:
> And how about the switch to reqmgr2? is it intentional ?

Yes.

geneguvo:
Hi Seangchan,

One more thing: I see you are setting REQMGR2 = "https://%s/reqmgr2" % HOST for all the workqueue flavors (dev/preprod/prod). Shouldn't it point to dev/preprod/prod URLs instead? just like the REQMGR URL...

ticoann:
> One more thing: I see you are setting REQMGR2 = "https://%s/reqmgr2" % HOST for all the workqueue flavors (dev/preprod/prod). Shouldn't it point to dev/preprod/prod URLs instead? just like the REQMGR URL...

Actually HOST there  was hard coded except the default flavor. (I think a little confusing since you mentioned) 
https://github.com/ticoann/deployment/blob/reqmgr2_transition/workqueue/GlobalWorkQueueConfig-prod.py#L18

geneguvo:
Ok. Sorry for the buzz. I'll adjust the other URLs to be based on HOST as well to avoid confusion.

On a side note: why  "config.WorkQueueManager.queueParams['PhEDExEndpoint'] = PHEDEX" is only present in the workqueue-testbed configuration? is  this something only used in testbed or was forgotten to be cleaned up? may I delete it?

ticoann:
> On a side note: why "config.WorkQueueManager.queueParams['PhEDExEndpoint'] = PHEDEX" is only present in the workqueue-testbed configuration? is this something only used in testbed or was forgotten to be cleaned up? may I delete it?

I didn't even realize that is there. I seem it is added there since default value is set to production PhEDEx.  So it should be there.
It would be more clear to have others instance as well. I will update that.

Thank you very much for cleaning up my mess.

PR #174, opened 2015-04-02
change HC script run time

altundag:
- for details, you can visit the SSB scripts twiki page:
    - https://twiki.cern.ch/twiki/bin/view/CMSPublic/CmsSsbScripts


geneguvo:
Thanks Mehmet! I've just applied this change on the machine.

PR #176, opened 2015-04-10
add logdb deployment

ticoann:
Tested in my vm. Please review.

### DIFF
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -11,6 +11,8 @@ from os import path
 HOST = socket.gethostname().lower()
 BASE_URL = "@@BASE_URL@@"
 DBS_INS = "@@DBS_INS@@"
+COUCH_URL = "%s/couchdb" % BASE_URL
+LOG_DB_URL = "%s/wmstats_logdb" % COUCH_URL
 
 ROOTDIR = __file__.rsplit('/', 3)[0]
 config = Configuration()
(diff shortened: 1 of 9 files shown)


geneguvo:
I have not tested it yet, but it looks good!

On your request in TRAC, however, you mention "Logdb contains 2 couchdb", but I see only "wmstats_logdb". For T0, there's "t0_logdb" and "t0_local_logdb". Is it intentional?

On a separate subject. What does this new thread and couch databases need in terms of requirements? I need at least the expected ~nr of requests per update cycle; expected ~nr of documents on the couch databases; amount of memory used by the thread when doing the update. Remember we had some issues with wmdatamining, so it is important we know the requirements in order to find/prepare the right back-end to run it.

ticoann:
> On your request in TRAC, however, you mention "Logdb contains 2 couchdb", but I see only "wmstats_logdb". For T0, there's "t0_logdb" and "t0_local_logdb". Is it intentional?

Sorry Diego, that was a mistake, in last update we remove local_logdb for both reqmon and t0_reqmon but I forgot include t0_reqmon changes.  Patch is updated now.

In terms of size and memory usage, right now it is pretty negligible. It only update 2 documents under 1K size per document per request for only active requests. Since it is the error case not all request will be updated. All the document is update not newly create since we compact every day the db size will be < 1M.  Memory used by new thread should be minimal as well  < 1M. And 7 day or older data will be deleted. (Although deleting policy will be changed)

However, we will add a lot more log for next deployment.  I am not sure how it will grow but rough estimate will be (generous max value)
1. number of documents in couch: 100K (~1KB per document)
2. update rate (not from the thread in cmsweb but from other resources wmagent). ~100 per min but it will be sparse there will be pick and no update, etc. 
3. Memory usage in new thread: This should be still minimal only used for deletion by request. (100 keys) ~1M.

I can give you more accurate when we start adding things. 
Thanks,

geneguvo:
Hey Seangchan,
In the meantime we've pushed the changes already. Would you please open a new PR with the updates  only? thanks.

ticoann:
Hi Diego, it seems the change is already pushed. (no t0_local_logdb)
https://github.com/dmwm/deployment/blob/master/t0_reqmon/deploy
Thanks,

PR #179, opened 2015-04-21
Upgrade of Castor agent for slc6

vanbesien:
@Diego: We can look at it in one of the coming days


geneguvo:
This one doesn't need to be updated/resubmitted as I already managed to extract the needed changes from it.

PR #185, opened 2015-05-01
Allow user to specify MYSQL socket file or use default

### LINKED ISSUE #184: Be able to specify the MySQL socket filename
Socket filenames (with path) in linux can be at most 117 characters long. Burying the socket file deep in the deployment can blow this limit. Instead, allow the socket file to be set by an optional environment variable.

ericvaandering:
Fixes #184. @ticoann @geneguvo please review

### DIFF
--- a/wmagent/manage
+++ b/wmagent/manage
@@ -42,6 +42,11 @@ ORACLE_TNS=
 
 MYSQL_USER=
 MYSQL_PASS=
+if [ -z "$DBSOCK" ]; then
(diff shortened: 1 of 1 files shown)


ticoann:
Hi Eric, So DBSOCK will be separately, right? I mean if someone want to use DBSOCK, he has to set the env valuable by himself.

ericvaandering:
Yes, exactly. I've tested this and it works just fine, but by default nothing changes.

geneguvo:
It looks fine to me.

ticoann:
Diego, could you merge this?

PR #188, opened 2015-05-12
pass couch port from secrete file

ticoann:
@ericvaandering, Eric, please review. This sets the couch port from secrete file not hard coded value.


wmagent/manage
```diff
@@ -207,7 +208,8 @@ load_secrets_file(){
         echo "Secrets file doesnt contain COUCH_PASS";
         exit 1
     fi
-
+    
+    COUCH_PORT=${MATCH_COUCH_PORT:-$COUCH_PORT};
```
ericvaandering: I don't know my bash. is colon-dash some way of doing "or take this value if not defined?"

giffels: Hi Eric, exactly.
http://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

ericvaandering:
Anyhow, this PR looks fine to me.

ticoann:
@geneguvo, please review and merge this. Tested in my vm.

PR #189, opened 2015-05-13
Changes for DCAFPilot project

vkuznet:


geneguvo:
Hi Valentin,
I've been reviewing this service and managed to make various changes myself, which I'll soon push upstream to make our the followup easier.

One of the problems I have now is that the cherrypy configuration is not shipped together with the deployment scripts, but is hardcoded in a file in the RPM. Would you please move that to deployment? either assume config.py would be in PYTHONPATH (which we'll add in the manage script), or make it a command line argument to bin/dcaf_server.

Other problem I noticed is that after having run the cronjobs, some of them are returning errors (models and verify). You can deploy it on your devvm and check them (once I push my changes upstream).

Yet another less important issue, the dataframe cronjob is generating a log under /data/srv/state/dcafpilot/analytics/data/log/, not /data/srv/log/dcafpilot/. I don't know if that was on purpose since there's already a log under  /data/srv/log/dcafpilot/cron4dataframe.log containing something.

We also need to talk about the requirements/resources needed for this application, if the plan is to deploy it on cmsweb.

Cheers,
Diego.

geneguvo:
I've pushed my changes. You can go ahead and deploy it on your devvm to see the errors. I'm reopening the PR so we can move forward with the discussion.

I forgot to say that the problem I had with the cherrypy configuration is that it is hardcoded to mount on "/", not on "/dcafpilot" as we'd do on cmsweb. Anyhow, if the configuration would be in a separate file under deployment I would do that change myself.

vkuznet:
Diego, thanks for review. I'll work with your commit in my VM. Regarding cherrypy config/mount point issue. Do you really want separate cherrypy config file or should I make configurable mount point? Right now I pass port/prediction area into dcaf server and I can add mount point as well. Otherwise I can put them into dcaf config file and init cherrypy config from them for dcaf server.

vkuznet:
Diego, which version should I use for dcafpilot deployment? Basically I need to know which cfg tag to use in step 7 of VM setup procedure.

geneguvo:
Use a separate cherrypy config file.

Since I've pushed it yesterday and there's no new release yet, you can use HEAD of deployment, and rpm VER to be the latest "comp" rpm version from http://cmsrep.cern.ch/cmssw/comp.pre/RPMS/slc6_amd64_gcc481/?C=M;O=D, which is "HG1507b-comp3".

vkuznet:
@geneguvo, Diego I made necessary changes for the package, i.e. I updated configuration file with new section for web server, it contains everything for cherrypy (port, mount point, etc.). The new version of DCAFPilot is 0.1.14 and I updated this PR with necessary changes. Please review.

vkuznet:
Diego,
I back from vacation and trying to catch up with open tasks. This issue was
pending in my queue for some time and I add configuration changes to DCAF
package. Could you please review them in this PR:
https://github.com/dmwm/deployment/pull/217

I combined web server configuration within main cfg file and use it in manage
script. I hope this is sufficient for all needs. This change will take place
with 0.1.15 tag I just applied to DCAFPilot package.

Once you review I'll perform tests with log/cronjob as you requested and
submit final tag for DCAFPilot package.

Thanks,
Valentin.

geneguvo:
Closing this PR. Let's follow up with this in the new PR you created, if anything would still need changes.

PR #194, opened 2015-05-27
couch 1.6 tier0

ticoann:
@hufnagel, Dirk please review.


hufnagel:
Deployment works, John or Luis will test it in the next replay and we can see if it fixes the replication issues.

PR #197, opened 2015-06-04
Update to popdb deploy script for version 0.3.3

nikmagini:
This removes from the popdbweb deploy script a parameter that is no longer used since version 0.3.3

Note that this is dependent on the corresponding pull request for popdb:

https://github.com/cms-sw/cmsdist/pull/1602


nikmagini:
Hi Bruno,

just to avoid misunderstandings: this change to the deploy script should not be applied until you update popdbweb to this version 0.3.3: https://github.com/cms-sw/cmsdist/pull/1602
The current version of popdbweb 0.3.2 won't work with this deploy script

PR #203, opened 2015-06-19
Fix mkdirs in inited_couch and activate_asyncstageout.

AndresTanasijczuk:


AndresTanasijczuk:
I think @dciangot will also test this PR (I have already tested it and everything seems fine). 
At the end @HassenRiahi should approve it.

PR #205, opened 2015-06-26
Style and layout patches for hg1507

vanbesien:
Two extra patches for hg1507.

@BrunoCoimbra I just pushed it to the same branch, like we do for cmssw. Let me know if you want me to do it differently.


BrunoCoimbra:
@vanbesien I didn't understand what you mean by pushing it to the same branch, but when I tried to apply the changes, most of them failed. Would you tell me how you made this changes, so we can figure out how to push them?

vanbesien:
Yeah, kinda makes sense, since the first part was already merged. I'll quickly just create a new branch and new pull request.

PR #208, opened 2015-07-02
Prep dqmgui prod switch to slc6

vanbesien:
For the config script:
- turning on acrontab entries for zipbackup, zipbackupcheck and indexbackup for new SLC6 servers
- on the old SLC5 servers, the stageout will have to be turned of manually

For the manage script:
- ZipDaemon and ZipFreezeDaemon will now run on the new SLC6 servers.
- on the old SLC5 servers, both Daemons will have to be removed manually from manage
- On the test server: Changed freeze period for zip files from 1 to 7 days, like on the other servers.

### DIFF
--- a/dqmgui/deploy
+++ b/dqmgui/deploy
@@ -66,7 +66,7 @@ deploy_dqmgui_post()
      echo "17 2 * * * $cmd"
      # Set the castor stageout (root files backup) on slc5 machines: run by cron under cmsweb account
      case $host in
-       vocms013[89] )
+       vocms13[89] )
        echo "@reboot $project_config/manage xstart stageout 'I did read documentation'" ;;
      esac
    }) | crontab -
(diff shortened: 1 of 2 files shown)


dqmgui/deploy
```diff
@@ -93,26 +93,23 @@ deploy_dqmgui_post()
        esac
 
        # Set the backup of the index (index CASTOR stageout)
-       # Currently only on SLC6 dev machine
-       # After the parallel run this must be switched on on the prod machines
+       # Only for Offline or Relval SLC6 machines
        case $host in
-         vocms0133 | broen* )
+         vocms013[389] | broen* )
```
rovere: Just a minor point: private VM names should not enter production deployment scripts...

vanbesien: Indeed. Issue was discussed before. Given that we were making quite some changes to the manage and config scripts, this became at some point the most convenient way to test them without cheating.
It's on my planning to remove it again.

rovere: +1.

PR #215, opened 2015-07-31
Post SLC6 migration Offline DQM GUI cleanup

vanbesien:
A collection of changes we wanted to make after the migration of Offline DQM GUI to SLC6.
Amongst many commits, the most important are:
* Removal of everything CAF related
* Addition of 2 new servers to the test configuration on vocms0131

@geneguvo and @BrunoCoimbra commit 9bf16c1 makes quite a lot of changes to the frontend configuration. Don't hesitate to take a look at it. And if not, most of this will be testable combining the testbed frontend with the first test deployment on vocms0131.

@rovere This PR also contain the switch in the backup location for the zip files in Castor: cc4ae83

Note that this is not the end, as it doesn't contain any of the updated for the Online SLC6 migration. This will come soon.


geneguvo:
Hey Broen,

From what I see on the changes, we're not going to run anymore the zipbackup and indexbackup on the cmsweb-testbed machine (vocms0131). I suppose this was intentional, right? anyway, as long as you have some reliable way of testing those components before we have to push the release to production servers, I'm fine with that.

Cheers,
Diego.

vanbesien:
Hey Diego,
Indeed. The zip backup chain was not there previously either. I added it some months ago to start testing all Castor-related development. I consider this finished and there's no point in constantly pumping useless data to Castor. When new development happens, we can simply add it again during the test phase.

vanbesien:
@geneguvo I see that since today there are merge conflicts for this PR, but Github doesn't show me in which files. I assume it's on one/some of the frontend/* files. How do you want to proceed?

geneguvo:
I'll solve the conflicts when merging this PR with the other changes I pushed this morning. Don't need to worry about that.

Concerning the castor stuff, it is important it gets tested there somehow, even if syncing only a subset of data or cleaning it more often. It is not only about the dqmgui development of the components themselves, but problems might come due to the new versions of castor clients (which we take from the system), changes to the machine configuration (i.e. firewall openings), and even changes on the CERN/IT side as it happened ~1.5 year ago when they closed anonymous access to it, or when they set access through the archive pool only.

Anyway, this is an important part of the data flow that needs to be validated, and I don't think space on castor is an issue for it. So unless it is not generating any other trouble or mess, let's keep it. In the old servers the index backup acronjob was running on testbed weekly, while monthly in production servers.

Cheers,
Diego.

vanbesien:
Ok, in fact, after giving it some more thought, I totally agree.
I'd still prefer you to merge this PR the way it is now. Is that okay?

I'll give it some thought in the coming weeks, do some data rate measurements so that we can make an educated choice between constantly backing up all dev data or maybe putting some filter to reduce it.
I'll make a new PR by the end of the month then to handle this.

geneguvo:
Yes, that's ok.

One minor thing I've noticed is that on backends-dev.txt you have set dqm/dev to point to vocms0131, while it should point to the local instance of dqmgui. That's the mapping that is used when we deploy on devvms or on cmsweb-dev. If we point it permanently to 0131 there's no way to reach the local running instance through the frontend, so we cannot make any sanity checks. I'll change it to point to localhost, but I'm not sure I should do the same for offline-test|relval-test. Does it make sense to have offline-test and relval-test running on my private devvm deployment too?

vanbesien:
Hey Diego,

Your point makes total sense. I simply didn't realize this while changing the files. I'd suggest to change it to localhost indeed and remove the |offline-test|relval-test.

geneguvo:
Since offline-test|relval-test only exist on vocms0131, I left them there and set only "dev" to go to localhost. I'll test everything (my changes + the merge) tomorrow morning. If all goes fine I'll push it upstream. Cheers, Diego.

PR #216, opened 2015-07-31
Update schedd monitor redirections ids

mmascher:


mmascher:
Hold on, I also need to modify the rewrite rule (it's matching just one digit [0-9], it needs to match more)

PR #218, opened 2015-08-03
change ES layouts

cmkuo:
Add ES recruit energy spectrum with selected good rechits to top layer of ES DQM


vanbesien:
@geneguvo @BrunoCoimbra This PR can be simply closed. Was superseded by 223 which we are testing now.

PR #224, opened 2015-08-09
Hcal Plugin Updates

vkhristenko:
An attempt to provide new Plugins for HCAL DQM.
In this PR 2 things added:
HcalCalibRenderPlugin.cc - Render Plugin for HcalCalib Application. HcalCaliRenderPlugin::applies(...) will return true only if HcalCalib/ is found in the pathname of the ME....
HcalObjectCustomizer.h - Generic Hcal TObject Customizer.

The question of Layouts has been discussed with Fede/Dan. At this point HCAL DQM in the state of transition and after this PR Layouts will be outdated as these plots will no longer exist! 
NOTE: Above Sentence only applies to HcalCalib Application

VK

### DIFF
--- a/dqmgui/style/HcalCalibRenderPlugin.cc
+++ b/dqmgui/style/HcalCalibRenderPlugin.cc
@@ -1,726 +1,128 @@
-/*!
-  \file HcalRenderPlugin.cc
-  \brief Display Plugin for Hcal DQM Histograms
-  \author J. Temple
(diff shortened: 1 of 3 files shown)


vanbesien:
@BrunoCoimbra @geneguvo 
This pr has been tested on the Online and is already in production there.
So, please merge.

Thanks!

PR #227, opened 2015-08-12
Offline test config changes

vanbesien:
Config update for offline-test and relval-test
 Added the Castor backup back to the testing server (without any throttling on the data stream)


vanbesien:
I just added a very small commit for the online server config as well.

For me this is ready to be merged.

PR #230, opened 2015-08-18
Fixed bugs related to postDraw feature that was not previously used b…

dinardo:
Fixed bugs related to postDraw feature that was not previously used by this app.
Basically I've moved some settings from preDraw to postDraw.
I've also added two histograms to monitor the failure mode bits of the app.
This patch is for the an application that, as far as I know, runs only on online DQM GUI.


vanbesien:
@BrunoCoimbra This patch was tested on the Online and can be merged.

PR #245, opened 2015-09-28
Adding colz options to some TH2 tracking plots

hdelanno:
A cosmetic update to add the colz option to TH2 plots for the tracking.


vanbesien:
@BrunoCoimbra @geneguvo Please merge.
Should be really easy to test on the offline-test GUI on 0133.
* If this can go in HG1510, that would be nice.
* If not, we can wait for HG1511, which should be fine too.

PR #248, opened 2015-10-05
Adding Plots to HCAL DQM GUI Layouts

vkhristenko:
Adding 4 plots for HCAL DQM GUI Layout.

I see that old commits from previous PR show up here, but do not affect the PR itself as those have been already integrated. Should be fine I think...

Let me know
VK


vanbesien:
Superseded by #249, please dismiss and close.

PR #250, opened 2015-10-07
Addition of the col z option for new TH2

hdelanno:
Addition of the col z option for new TH2


hdelanno:
Found a way to do this in a more efficient way. Will update the PR soon.

vanbesien:
@hdelanno I guess you can close this one. You should have a close button somewhere.

PR #252, opened 2015-10-08
Colz options added to more plots in the Tracking DQM

hdelanno:
This is an update of the PR #250 
It adds the requested colz option to 2D plots in a bit more efficient way than previously.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #256, opened 2015-10-16
Cleanup and documentation of the workspace configs

vanbesien:
The rationale behind it is that the configs are changed by normal users,
while the code that interprets it is hidden very well and very deeply,
even in a different repository. Adding documentation to a Twiki would be
another option, but it would be bad practice to try to document a
method's parameters in a Twiki. So we go for the least bad option of
duplicating the same info in the 4 config files.


vanbesien:
@BrunoCoimbra , @geneguvo
    Tests OK.
    Can you merge this?
    Thanks

PR #265, opened 2015-11-13
Update of the layouts for tracker

hdelanno:
Adding one plot in the shifter and online workspaces for SiStrip (HLT_Random_v2 counts)
Changin the z scale for one offline plot in Pixel for Endcaps Hit Efficiency


hdelanno:
Looking at the Online Test, I found some minor problems.
I fixed them.
Would it be possible to redo the tests?

Thanks !

hdelanno:
After a discussion with Tomas and Francesco, I close this PR.
For the moment we will just update the short term instructions and, in the future, I will make our own plot checking if a Random is trigged.

PR #279, opened 2016-01-21
Candidate release 1.2.9.pre.1

vinnie87:
More configuration fixes


fwyzard:
@BrunoCoimbra , how do we select which of the files that are in the repository are included in the RPM package that is deployed ?

PR #281, opened 2016-02-04
change the order of database schema initialization

hufnagel:
Change the order the database schema are initialized as the Tier0 schema now depends on the Database schema.


hufnagel:
Can this please be merged asap ?

PR #296, opened 2016-03-14
Cleared content of the L1T shifts layouts

vanbesien:
Both online and offline -> don't know why there was offline anyway.
Both L1T and L1TEMU
Folders will be hidden completely since there are no plots inside.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #299, opened 2016-03-16
Removed Castor from the shift layouts

vanbesien:
There are no instructions and as far as we know Castor will not be
in before the HI run at the end of the year. We leave the expert
workspace for what it is.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #301, opened 2016-03-21
Change time scale for splash

mbandrews:
**Timing scale changes for beam splash only:**

- Change timing map z-axis scale to +/- 25 ns
- Change timing projection y-axis scale to +/- 25 ns


vanbesien:
This one should NEVER be merged.
Only for online - only during splashes

vanbesien:
@mbandrews 
Hey Mike, can you close this? Thanks.

PR #303, opened 2016-04-01
Hid Castor and ZDC from the HV plot

vanbesien:
These detectors will not be in for quite a while and otherwise they
would always show up as "red".


vanbesien:
@BrunoCoimbra
Tests OK.
Can you merge this? (This is mainly for online, so not a disaster if it can not go in the current release)
Thanks

PR #309, opened 2016-04-12
Fixed the rainbow palette problem

vanbesien:
Previously pre-defined colors would be selected by their index and then
the color would be overwritten by SetRGB. People were using random high
numbers like 950 or 2000 as their start indexes.
However with the newest root updates, they actually started to overwrite
existing colors.
We replaced those methods to use GetColor(r,g,b) witch will create the
new color _outside_ of the range if it does not exist yet.


swang373:
Bravo!

vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #313, opened 2016-04-19
Filter timing and exit status from DQMGUI indexing

rovere:
Every time a file is indexed by the DQMGUI, the time spent
in the process and the exit status of the process itself are
logged. This PR updates the monitoring system to be able to
extract these informations and make them available later in
the chain.


rovere:
Can we have this included in the coming production release?

PR #322, opened 2016-04-22
Fix rendering of empty TProfile2D bins

mbandrews:
Fix rendering of empty bins in all TProfile2D plots so that they display white instead of purple.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #325, opened 2016-04-27
Enabled checkpoint for WMAgent; added replication performance tweaks

amaltaro:
Big thanks to @geneguvo who got this setup working!
By default, we set the checkpoint interval to 10min (600k milisecs).

The performance settings need to be carefully watched in the production nodes to make sure it won't bring bad effects.
@ticoann FYI


ticoann:
@amaltaro, Alan the production couch is already updated? Can this be applied current production agent?

amaltaro:
Production agents don't have these settings yet, however I'm pushing it to vocms0308 right now.

PR #326, opened 2016-05-02
Real vs emul TP Et layouts+rendering, and three new l1trigger plots

tanmaymudholkar:
1) Added RealvEmul TP Et layouts+rendering
2) Added three plots from the L1 Trigger DQM to ECal layouts, and updated ECal workspace filter to allow display


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #333, opened 2016-05-13
DT Layout: added residuals for 09-Hits and 10-Segments

jfernan2:


dqmgui/layouts/dt-layouts.py
```diff
@@ -109,6 +109,29 @@ def dtlayout(i, p, *rows): i["DT/Layouts/" + p] = DQMItem(layout=rows)
...
+	    dtlayout(dqmitems, name,[{ 'path': histoname}])
+
+#### SEGMENTS RESIDUALS #########################################################################
+for wheel in range(-2, 3):
+    for station in range (1, 5):
+        for sector in range (1, 15):
+            if station != 4 and (sector == 13 or sector == 14):
+                continue
+	    for sl in range (1, 4):
+	    	if sl == 2 and (sector == 13 or sector == 14):
```
vanbesien: Hey @jfernan2 
I see a horrible mix of tabs and spaces here. Python will only allow this code to run, because it assumes that your tabs are equal to 8 spaces, otherwise it would give an indentation error because lines 126 and 128 are not alligned.
Git conveniently show tabs as 7 spaces, and this is why you see the indentation problem between line 126 and 128 or between 116 and 118.
Line 129 starts with a tab, followed by 4 spaces, followed by a tab and then the code starts. This is obviously not okay.
Please clean this code. Preferably don't use tabs at all and use only spaces, since you can make no assumptions on how wide your tabs are for any other editor or interpreter.
Contact us if you have troubles.
Thanks

jfernan2:
Hi @vanbesien 
I have fixed the issue, I'm sorry for the incovenience; I was not aware of this problem
Thanks

jfernan2:
Hi Broen,

sorry, I spotted a last minute bug from my side which I had not seen in my test DQM GUI due to lack of statistics, I have pushed the right commit into the PR, now I have ensured it is OK

I'm sorry once again for the incovenience

vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #334, opened 2016-05-14
Add di-electron plot from HLT

mbandrews:
**Add di-electron mass monitoring from HLT to check HLT e/gamma rates**
- ECAL was recently implicated in the problem of missing e/gamme rates in HLT which turned out to be due to a bug in the HLT code (nothing peculiar was seen in ECAL DQM)
- The di-electron mass plot is primarily fed by the e/gamma HLT path and, for future reference, ECAL would like to be able to conveniently monitor this plot as well


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #338, opened 2016-05-25
Add SR flag mismatch plots and fix EB label numbering

tanmaymudholkar:
Added SR flag mismatch plots to online Ecal layout in Ecal/07 Selective Readout, and also modified corresponding layout plugin. Also, fixed EB TT numbering, now consistent with other plots.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #344, opened 2016-06-13
Fixed broken FED error plot

vanbesien:
For ages already the RPC shift and expert layouts show that the FED
error plot (01-Fatal_FED_Errors) is not available. The reason is simple:
It's linking to the wrong directory: FEDIntegrity_SM instead of
FEDIntegrity_EvF. The fix is so simple that I'm doing it myself.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #346, opened 2016-06-21
Add noise correlation+difference plots

mbandrews:
- Add noise correlation and difference plots: display with under- and over-flow in stat box (include obj->SetStats(true) in postDrawbyName)
- Modify some descriptions of SRP desync errors
- Return TTFlagMismatch to default color palette


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #355, opened 2016-07-19
Phase1 Pixel DQM Renderplugin changes

schneiml:
This PR prepares the SiPixel Renderplugin for Phase1,  by:

- Removing lots of code that was probably never used.
- Applying more options by default (so also for new Phase1 plots)
- Providing a way to get Layer/Shell/etc. markers on profiles. The information has to be provided in the axis label string. The code for this is not yet PR'd against CMSSW, since it makes the labels hard to read w/o this PR. This got much more complicated than expected, but the code should now be clean and even work properly.
- Some more general style changes that also apply to Phase0, related to font sizes and axis ranges.
- The renderplugin now puts the full pathname of the ME on the canvas, to prevent confusion of plots.

### DIFF
--- a/dqmgui/style/SiPixelRenderPlugin.cc
+++ b/dqmgui/style/SiPixelRenderPlugin.cc
@@ -117,16 +117,28 @@ private:
 
   void putMarkers(TH1* obj) 
     {
+      // TODO: Y-Axis as well?
       TAxis* ax = obj->GetXaxis();
       auto label = std::string(ax->GetTitle());
       auto res = LabelMarkerParser<std::string::iterator>::parse_label(label.begin(), label.end());
-      if (!res.first ||  res.second.markers.size() == 0)  {
+      if (!res.first || res.second.markers.size() == 0)  {
         return; // parse failed, probably no markers
       }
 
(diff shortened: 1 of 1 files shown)


dqmgui/style/SiPixelRenderPlugin.cc
```diff
@@ -215,118 +203,108 @@ void preDrawTH2( TCanvas *, const VisDQMObject &o )
...
+      xa->SetLabelSize(0.03);
       ya->SetTitleOffset(0.75);
-      ya->SetTitleSize(0.065);
-      ya->SetLabelSize(0.065);
+      ya->SetTitleSize(0.04);
+      ya->SetLabelSize(0.03);
+      TGaxis::SetMaxDigits(3);
+
+      // Always include 0.
+      if( obj->GetMinimum() > 0.) obj->SetMinimum(0.);
```
schneiml: TODO: this causes trouble on logscale plots. I'll look into a fix. ( @vanbesien )

vanbesien:
@BrunoCoimbra , @geneguvo
Can you merge this?
Thanks

PR #365, opened 2016-08-16
Removed online workspaces that are not maintained

vanbesien:
In practice: the SiStripLAS workspace of which nobody knows what it is
and that only contains broken plots.
Also all the "feedback for collisions" workspaces, since they are
undocumented and mostly broken. We contacted all the different
subsystems to find out if anyone knew the purpose or would be able to
update them if requested and the response was 100% negative.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #366, opened 2016-08-16
Adding plugin .h-files to list of files to compile

vanbesien:
This is step 2 of a 2 step change:

Step1: Make sure that _if_ .h files are taken into account for the
decision on whether or not something has changed, then these .h files
are not sent as input to the make file.

Step2: Once this is done, we can add the header files to the patterns of
the files that need to be looked at for the render plugins. (This is
done in configuration, which is a different repo.)


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #368, opened 2016-08-22
PhedexReplicaMonitoring deployment

vkuznet:


phedexreplicamonitoring/manage
```diff
@@ -0,0 +1,168 @@
...
+{
+  echo "status $ME"
+}
+
+# Runs actions to be done through cronjobs
+cron()
+{
+  local action=$1
+  local fromdate=`date +%Y-%m-%d -d "yesterday"` # GNU date utility
+  local todate=`date +%Y-%m-%d`
```
nikmagini: IIRC to get a one-day interval it should be
local todate=$fromdate

PR #369, opened 2016-08-23
add a HCAL DQM role for host hcal-dqm-fu (fu-c1f11-21-03)

deguio:
@vkhristenko


vanbesien:
@BrunoCoimbra @geneguvo This PR will only affect the new online DQM GUI server that HCAL is adding. Please merge so that they can use it without having to do manual patches.

rovere:
Didn't we agree not to touch anything related to central configurations??

BrunoCoimbra:
So, just let me know if I should merge it or not! :)

vanbesien:
We (Marco and I) discussed a bit and came up with a cunning plan.
For now we would like to have it merged.

@vkhristenko @deguio This way you can easily keep your GUI updated. We count on you to keep it in sync with the production GUIs by upgrading on a regular basis.

deguio:
sounds good.
thanks,
F.

PR #378, opened 2016-09-05
add WMArchive URL

hufnagel:
@ticoann , if you make changes to the wmagent/manage script that are mandatory, can you please make them to the tier0/manage script as well (or at least let me know) ? Got surprised by this (without it we can't use any 1.0.20 WMCore tags).


ticoann:
@hufnagel, I am sorry Dirk, I didn't realize Tier0 case. I will be more careful about that.

PR #383, opened 2016-09-25
Offline Layouts update for HCAL DQM

vkhristenko:
Updated offline layouts for hcaldqm.

@deguio


vanbesien:
@BrunoCoimbra , @geneguvo
This is one for the offline GUI. Our code checks and smoke tests are ok, so please go ahead and merge this so that it can be tested on offline-test.
Thanks

PR #388, opened 2016-10-11
change range of zero occupancy trend plots to reflect Run2 data

fioriNTU:
Change axis range to relevant plots for this PR https://github.com/cms-sw/cmssw/pull/16154 in a way they reflect the number of zero occupancy rocs seen in Run2
Removing or retuning misleading blue lines on few plots in Pixel and Strip


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #395, opened 2016-10-23
moved default log scale for y-axis from postDraw to preDraw

idebruyn:
For a few plots in the SiStrip workspace. Default will still be log scale, but now it can also be changed to linear in the GUI. This PR affects both online and offline DQM.


vanbesien:
@BrunoCoimbra , @geneguvo
Tests OK.
Can you merge this?
Thanks

PR #402, opened 2016-11-10
Phase1+PilotBlade Online Trends Render Plugin

schneiml:
This PR adds a new Render Plugin that implements an experimental feature for the online GUI that is especially meant for timing scans and should also be used with the Pilot Blade.

The idea here is to collect a trend over time in a 2D histogram with "time" on the y-axis (here measured in blocks, which are an arbitrary multiple of lumisections), and render this 2D histogram as a bunch of overlaid, normalized 1D histograms.

The Render Plugin should not apply to anything in current DQM, so I think it is save to add to the GUI, though it will only be needed in special online setups. For production use, the memory management should be checked by somebody more knowledgeable in the ROOT/dqmgui details involved here.

Some of the related cmssw code is here: https://github.com/cms-sw/cmssw/pull/16482 , but the final setup for this sort of plot is only availble in the Pilot Blade branch (https://github.com/schneiml/cmssw/tree/phase1pixeldqm-pre9-pilot-with-pre16-changes).

(@threus @vanbesien @dmitrijus @fioriNTU @hdelanno

### DIFF
--- a/dqmgui/style/SiPixelPhase1OnlineRenderPlugin.cc
+++ b/dqmgui/style/SiPixelPhase1OnlineRenderPlugin.cc
@@ -1,6 +1,40 @@
 /*!
   \file SiPixelPhase1OnlineRenderPlugin
   \brief RenderPlugin for histograms showing some history in online
+
+This Render Plugin was originally desinged for Pixel timing scans. It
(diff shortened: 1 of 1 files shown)


vanbesien:
@BrunoCoimbra
Tests OK.
Can you merge this?
Thanks

PR #409, opened 2016-11-21
Split DBS instances into six servers.  Create new cpnfig files, update rewrite roles, deploy and manage scripts.

yuyiguo:
Split DBS instances into six servers.  Create new cpnfig files, update rewrite roles, deploy and manage scripts.


BrunoCoimbra:
Merged #410.

PR #410, opened 2016-11-22
split DBS instance into six servers.

yuyiguo:


BrunoCoimbra:
Hi @yuyiguo, should I take this PR instead of #409?

BrunoCoimbra:
Ok, I've just saw your comment on GitLab!

PR #413, opened 2016-11-25
Fix a problem moving reqmgr couchapps to reqmgr2

amaltaro:
Complement this PR #412 
In which I was writing the couchapp commands into the wrong file.

BTW, during the deployment we should remove this file
`/data/srv/state/couchdb/stagingarea/reqmgr
otherwise couch manage script will try to source it (with the old couchapps commands), though there is no longer *reqmgr directory.

Bruno, I'm still testing it.


amaltaro:
@ticoann FYI

PR #417, opened 2016-12-01
Re-arranging workspaces and adding info workspaces

vanbesien:
For both the online and the offline GUI the workspaces are re-ordered
and made more logical and consistent.
We also added layouts for the "Info" plots, so that we could add
documentation. These layouts were then added as workspaces as well, so
that from now on, you can double-click on the HV plot in the summary
page to get more general info on the run, also allowing you to zoom in
to this plot.


BrunoCoimbra:
Should I merge it? :)

vanbesien:
@BrunoCoimbra
Tests OK.
Can you merge this?
Thanks

PR #418, opened 2016-12-02
SiStrip: Small improvements for readability of several plots

lmoreels:
- Good detector fraction report maps for error messages separately (in SiStrip/MechanicalView)
Used a pre-defined palette for the plots such that the colour differences are more pronounced for small deviations from 1. This can further be improved by a dedicated palette to give even larger colour variations in the range 0.85 to 1. (Values smaller than that are rare.) 
Note that there are no changes to the plots that the shifters have to monitor. A reference to these plots has been added to the SiStripOfflineDQMHistoDescription twiki.
- FED error per FED ID plot (in QuickCollection)
Added grey hashed lines to indicate the partition boundaries. Attached labels to indicate which partition is which.

### DIFF
--- a/dqmgui/style/SiStripRenderPlugin.cc
+++ b/dqmgui/style/SiStripRenderPlugin.cc
@@ -151,6 +151,12 @@ private:
       {
         obj->SetStats( kFALSE );
         dqm::utils::reportSummaryMapPalette(obj);
+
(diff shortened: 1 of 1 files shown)


vanbesien:
@BrunoCoimbra
Tests OK.
Can you merge this?
Thanks

PR #424, opened 2016-12-27
change to wq cherrypy thread

ticoann:
https://github.com/dmwm/WMCore/pull/7496 has to be merged first


ticoann:
@amaltaro, Alan please review. this is tested with dmwm/WMCore#7496

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+config.WorkQueueManager.couchurl = COUCH_URL
+config.WorkQueueManager.dbname = workqueueDBName
+config.WorkQueueManager.inboxDatabase = workqueueInboxDbName
+
+main = config.section_("main")
+srv = main.section_("server")
+srv.thread_pool = 30
+main.application = "workqueueserver"
+main.application_dir = "workqueue"
+main.port = 8250 # main application port it listens on
```
amaltaro: This port seems to be allocated to DBS3. We need to ask @BrunoCoimbra for a port range to be used here.

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+# Defaults to allow any CMS authenticated user. Write APIs should require
+# additional roles in SiteDB (i.e. "Admin" role for the "ReqMgr" group)
+main.authz_defaults = {"role": None, "group": None, "site": None}
+
+sec = main.section_("tools").section_("cms_auth")
+sec.key_file = "%s/auth/wmcore-auth/header-auth-key" % ROOTDIR
+
+# this is where the application will be mounted, where the REST API
+# is reachable and this features in CMS web frontend rewrite rules
+app = config.section_(main.application) # string containing "reqmgr2"
```
amaltaro: containing "workqueueserver"

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+workqueueInboxDbName = 'workqueue_inbox'
+wmstatDBName = 'wmstats'
+reqmgrCouchDB = "reqmgr_workload_cache"
+
+HOST = socket.gethostname().lower()
+BASE_URL = "@@BASE_URL@@"
+DBS_INS = "@@DBS_INS@@"
+COUCH_URL = "%s/couchdb" % BASE_URL
+
+REQMGR2 = "https://%s/reqmgr2" % HOST
```
amaltaro: Shouldn't this point to the cmsweb(-testbed) url instead of the local hostname?

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+    config.queueParams['RequestDBURL'] = "%s/%s" % (COUCH_URL, reqmgrCouchDB)
+    config.queueParams['central_logdb_url'] = LOG_DB_URL
+    config.queueParams['log_reporter'] = LOG_REPORTER
+    
+    config.reqMgrConfig = {}
+    config.reqMgrConfig['reqmgr2_endpoint'] = REQMGR2
+    config.reqMgrConfig['central_logdb_url'] = LOG_DB_URL
+    config.reqMgrConfig['log_reporter'] = LOG_REPORTER
+
+# Production instance of wmdatamining, must be a production back-end
```
amaltaro: "Production and pre-production instances for workqueue"

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+if  HOST.startswith("vocms0140") or HOST.startswith("vocms0131"):
+    extentions = config.section_("extensions")
+    reqmgrInteraction = extentions.section_("reqmgrInteraction")
+    reqmgrInteraction.object = "WMCore.GlobalWorkQueue.CherryPyThreads.ReqMgrInteractionTask.ReqMgrInteractionTask"
+    setWorkQueueCommonConfig(reqmgrInteraction)
+    reqmgrInteraction.interactDuration = 60 * 2 # every 2 minutes
```
amaltaro: It's currently set to 20min. I think we should initially keep the same polling cycle (I also doubt 2 min would be enough to acquire work).

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+    locationUpdateTask.object = "WMCore.GlobalWorkQueue.CherryPyThreads.LocationUpdateTask.LocationUpdateTask"
+    setWorkQueueCommonConfig(locationUpdateTask)
+    locationUpdateTask.locationUpdateDuration = 60 * 5 # every 5 minutes
+    locationUpdateTask.log_file = '%s/logs/workqueue/locationUpdateTask-%s.log' % (__file__.rsplit('/', 4)[0], time.strftime("%Y%m%d"))
+    
+    # workqueue cleanup threads
+    cleanUpTask = extentions.section_("cleanUpTask")
+    cleanUpTask.object = "WMCore.GlobalWorkQueue.CherryPyThreads.CleanUpTask.CleanUpTask"
+    setWorkQueueCommonConfig(cleanUpTask)
+    cleanUpTask.cleanUpDuration = 60 * 5 # every 5 minutes
```
amaltaro: Is this thread responsible for moving requests to completed and cleaning all the GQ elements?
Current setting is 20min, maybe we set it to 15 or 25min to avoid having these threads running all the time in parallel(?)

workqueue/deploy
```diff
@@ -22,47 +25,32 @@ deploy_workqueue_sw()
...
+    dev )     base_url="https://cmsweb-dev.cern.ch"     ;;
+    * )       base_url="https://`hostname -f`"          ;;
+  esac
+  perl -p -i -e "s{\"\@\@BASE_URL\@\@\"}{\"$base_url\"}g;" \
+       $root/$cfgversion/config/$project/config.py
 }
 
 deploy_workqueue_post()
 {
+  case $host in vocms013[89] ) disable;;  * ) enable;; esac
```
amaltaro: We want to have these threads running in a single instance per cluster, thus I'd keep the previous configuration e.g.
`prod:vocms0140 | preprod:vocms0131 | dev:vocms0127 | dev:vocms0126 | default:* )`

workqueue/manage
```diff
@@ -0,0 +1,148 @@
...
+if [ -e $AUTHDIR/dmwm-service-cert.pem ] && [ -e $AUTHDIR/dmwm-service-key.pem ]; then
+  export X509_USER_CERT=$AUTHDIR/dmwm-service-cert.pem
+  export X509_USER_KEY=$AUTHDIR/dmwm-service-key.pem
+fi
+
+
+# Start service conditionally on crond restart.
+sysboot()
+{
+  LD_PRELOAD=$JEMALLOC_ROOT/lib/libjemalloc.so wmc-httpd -v -d $STATEDIR -l "|rotatelogs $LOGDIR/workqueue-%Y%m%d.log 86400" $CFGFILE
```
amaltaro: Just out of curiosity. I think this log will be mostly empty, no? Since all the workqueue activity will be written in the other 3 logs.

ticoann:
Hi Alan, Thank you very much for the review. I updated the code. 
There are still too things need to be updated/discussed - port number and server running backend before it gets merged

workqueue/manage
```diff
@@ -0,0 +1,148 @@
+#!/bin/sh
```
amaltaro: @ticoann it seems you need to make it executable.

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+    config.queueParams['central_logdb_url'] = LOG_DB_URL
+    config.queueParams['log_reporter'] = LOG_REPORTER
+    
+    config.reqMgrConfig = {}
+    config.reqMgrConfig['reqmgr2_endpoint'] = REQMGR2
+    config.reqMgrConfig['central_logdb_url'] = LOG_DB_URL
+    config.reqMgrConfig['log_reporter'] = LOG_REPORTER
+
+# Production instance of globalworkqueue, must be a production back-end
+if  HOST.startswith("vocms0140") or HOST.startswith("vocms0131"):
```
amaltaro: I still don't know how to handle it. But this block should be enabled for private VMs as well.

workqueue/config.py
```diff
@@ -0,0 +1,104 @@
...
+workqueueInboxDbName = 'workqueue_inbox'
+wmstatDBName = 'wmstats'
+reqmgrCouchDB = "reqmgr_workload_cache"
+
+HOST = socket.gethostname().lower()
+BASE_URL = "@@BASE_URL@@"
+DBS_INS = "@@DBS_INS@@"
+COUCH_URL = "%s/couchdb" % BASE_URL
+
+REQMGR2 = "https://%s/reqmgr2" % BASE_URL
```
amaltaro: Seangchan, you can remove `https://` here, since BASE_URL already has it (that's why WQ was not acquiring work in my VM :-)).

workqueue/deploy
amaltaro: @ticoann don't forget updating these hosts, otherwise we'll end up with these threads running in different backends and then the chaos is in place :)

amaltaro: Or maybe not, I've seen your `if HOST` down there. Which means, the service is enabled but it's actually not running any tasks (just a process listening on that port, isn't it?)

amaltaro:
@ticoann before it gets merged, can you please add the configuration I suggested in this PR:
https://github.com/dmwm/WMCore/pull/7531

? for the monitoring thread. Thanks

workqueue/deploy
BrunoCoimbra: To make it run on the machines Alan mentioned, I'd change this line as the following:
`case $host in vocms013[2689] vocms014[12] vocms016[135] ) disable;;  * ) enable;; esac`

ticoann:
Alan, I updated config. Could you check? You want to run this every 30 min, right?

workqueue/manage
amaltaro: Maybe it was how I applied this patch then that messed up with the permissions

workqueue/config.py
amaltaro: I just talked to Bruno, you can enable these threads for cmsweb-dev as well (vocms0127).
I agree about not having VMs here, let's just make sure Alex and others are aware that they need to manually enable workqueue on their VMs.

ticoann:
I added dev instance. (vocms0127). I will run the more test with deployment script.

amaltaro:
Thanks, Seangchan. These review got a bit weird, I got duplicated comments and some I cannot reply. If I missed any of your questions, just ask again please :)

About the server running on the general purpose backends or not, maybe we keep it only in vocms0140 for now and once we start making APIs available, then we migrate it to all the cmsweb backends. But honestly, I have no strong arguments otherwise.

amaltaro:
@ticoann , Bruno and I applied one line change to the workqueue post code, in order to blacklist (disable) all nodes but those 3 (prod,preprod,dev) nodes as we discussed before.
It will also enable this service in our VMs.

PR #425, opened 2017-01-06
remove reqmgr from wmagent deployment

ticoann:
@amaltaro, Alan please review, still need to be tested


amaltaro:
It seems we can still delete some more reqmgr stuff. Basically all reqmgr lines that are not `REQMGR2_URL` or `reqmgr2_url`.

PR #427, opened 2017-01-20
Phase1 DQM PixelMaps

schneiml:
This PR adds a new RenderPlugin that applies @jkarancs code to draw ROC and module boundaries on top of the Pixel Map plots in the new Pixel Phase1 DQM.

Those plots are defined using the SIPixelCoordinates class, which is actually more generic (not limited to Phase1, and not limited to DQM), so this plugin is also in some way more generic. However, it is currently only used in Pixel Phase1 DQM (versions newer than 900pre3, which is not out yet).

I hope the `l->SetBit(kCanDelete);` is enough to make this not leak memory.

Seems to work as expectec, no extended testing done so far.

@vanbesien @fioriNTU @hdelanno @enochnotsocool

### DIFF
--- a/dqmgui/style/SiPixelMapsRenderPlugin.cc
+++ b/dqmgui/style/SiPixelMapsRenderPlugin.cc
@@ -1,0 +1,250 @@
+/*!
+  \file SiPixelMapsRenderPlugin
+  \brief RenderPlugin for plots based on SiPixelCoordinates
+
+This RenderPlugin adds lines and markers about the ROC layout of Pixel modules
+to 2D plots with a per-ROC granularity. Such plots can be created using the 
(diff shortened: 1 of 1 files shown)


vanbesien:
I guess this is for both online and offline, right?

fioriNTU:
@vanbesien yes, it for both online and offline

schneiml:
@vanbesien as Franceso says, both offline and online. More precisely, the first occassion to test in production would be the upcoming 900pre3 relval. RelVal is considered more offline, I think...

vanbesien:
Frozen

vanbesien:
@BrunoCoimbra 
Can you merge this?

Would be great to go with the production release of 07/02. But if it can go in with the earlier-production-release with updated root, that would be even better.

Thanks

PR #430, opened 2017-01-26
add WMStats Erlang view

ticoann:
@amaltaro, Alan, please review. javascript view were clean up. so some of the view were broken.


amaltaro:
@ticoann ,
 @BrunoCoimbra is going to have a look at it because it seems we have legacy rules here. He'll fix it and create a new PR.

ticoann:
Thanks Bruno,

PR #433, opened 2017-01-29
Small layout style change ...

dinardo:
Small layout style change inline with author style ...

### DIFF
--- a/dqmgui/style/BeamPixelRenderPlugin.cc
+++ b/dqmgui/style/BeamPixelRenderPlugin.cc
@@ -1,4 +1,4 @@
-/*!
+/*
   \File BeamPixelRenderPlugin
   \Display Plugin for BeamSpot from Pixel-Vertices
   \author Mauro Dinardo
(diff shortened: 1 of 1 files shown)


threus:
@BrunoCoimbra
Can you please merge this?
Thanks,
Tomas for DQM

dinardo:
Hi,
forgive my naive comment, I might be to paranoid, but when do you think that this merging will occur ?
I just would like to avoid that it goes lost ... though I don't even know whether this can actually happen with git :-)

Cheers,
- Mauro.

vanbesien:
Not a very naive comment, in fact.
@BrunoCoimbra will you merge this? Or did we miss the deadline for the feb release?
@dinardo don't worry, since for the online, we manually patch the Online GUI and your patch is actually already deployed and active on the online DQM GUI.

BrunoCoimbra:
Sorry guys, I don't know exactly how, but it seems we've missed this PR.
Unfortunately now it's too late for February but I'll make sure it goes on the next release.

PR #434, opened 2017-01-30
Force IPv4 for dev DQMGUI server

rovere:
The underlying network layer of DQMGUI is forced to use IPv4
network addressing space since the DQMCollector application
is using such version. Usage of the generic 'localhost'
interface could result in IPv6 usage, depending on how the
hosting machine is configured, the facto preventing the Live
source to work properly.


vanbesien:
@BrunoCoimbra 
Please merge :-)

PR #438, opened 2017-02-10
fix tier0 time redirect view

ticoann:


amaltaro:
@ticoann I assume you also want to have this deployment change in the critical cmsweb fix we're planning for this week?

amaltaro:
Nevermind, he confirmed in a private mail that this is needed.

PR #449, opened 2017-03-15
Changing soundserver to new daq-expert.cms

vanbesien:
Changing this because DAQ people are replacing their daq-doctor by something called daq-expert


vanbesien:
@BrunoCoimbra , @h4d4 : Please merge

PR #450, opened 2017-03-16
Hiding CASTOR and ZDC from the HV plot again

vanbesien:
HI runs are over, so for the next 2 years CASTOR and ZDC will be off again. 
So we should not show a big red line anymore on the main HV summary plot when they are off, since that will be the normal situation anyway.


vanbesien:
@BrunoCoimbra @h4d4 
This can be merged right away

PR #458, opened 2017-03-29
Fix crash

fioriNTU:
Fixing crash on the rendering

### DIFF
--- a/dqmgui/style/SiPixelRenderPlugin.cc
+++ b/dqmgui/style/SiPixelRenderPlugin.cc
@@ -39,7 +39,7 @@ public:
     {
       c->cd();
       
-      if( dynamic_cast<TProfile*>( o.object ) || dynamic_cast<TProfile2D*>(o.object))
+      if( (dynamic_cast<TProfile*>( o.object ) || dynamic_cast<TProfile2D*>(o.object)) && o.name.find( "Lumisection" )!=std::string::npos)
       {
         TH1*  obj = dynamic_cast<TH1*>(o.object);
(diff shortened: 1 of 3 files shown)


dqmgui/style/SiPixelRenderPlugin.cc
```diff
@@ -269,7 +269,7 @@ class SiPixelRenderPlugin : public DQMRenderPlugin
 
       TH2F* obj2 = dynamic_cast<TH2F*>( o.object );
 
-      if( o.name.find("Summary") != std::string::npos && o.name.find("_Error_") == std::string::npos) 
+      if( o.name.find("Summary") != std::string::npos && o.name.find("_Error_") == std::string::npos && o.name.find("PixelPhase1/Summary")!=std::string::npos)
```
vanbesien: Hey, are you sure about this? You're basically removing the reportSummaryMapPalette for the original Summary with this line.

fioriNTU: right! I will fix now, I tend to forget about Phase 0, thanks for the notice

fioriNTU: Actually, it is not true, I have just checked, few lines below you can find the actual Phase 0 summary called "reportSummaryMap" that has the palette set correctly. I think that after my latest commit (that is not affecting online) this is ok.

threus:
@BrunoCoimbra @h4d4 could you please merge this PR?
Thanks,
Tomas for DQM

PR #465, opened 2017-04-11
Adding Tomas + Jim to receivers of agent warnings

vanbesien:
When the agents notice something weird they notify Broen and Marco with an email.
Usually we then wait to see if the problem persists and launch an investigation if needed.
We added two more people to the list: Tomas and Jim


vanbesien:
@h4d4 @BrunoCoimbra Please merge

PR #468, opened 2017-04-24
ROC indicator line / FED error bin labels

enochnotsocool:
1. Added visual indication of ROC boundaries in the per-module digi-occupancy plots. 
2. Added bin labels for different FED errors.
Due to the change in orientation and range of the plots involved, this change should only be implemented after the [CMSSW PR 18446]( https://github.com/cms-sw/cmssw/pull/18446) goes live.


threus:
@BrunoCoimbra @h4d4 could you please merge this PR?
Thanks,
Tomas for DQM

PR #471, opened 2017-04-26
(Temporary patch for splash events only) Changed axis range in timing plots.

tanmaymudholkar:
During splash events, we often have very poor signal shapes. In addition, it is useful to view these events in a wider time window than the default configuration on DQM. This PR changes the timing windows set in the render plugin. This PR is not meant to be merged into master.


tanmaymudholkar:
Closing this now because splashes are over.

PR #499, opened 2017-06-15
Summary Map back in layouts

fioriNTU:
This PR puts back the Summary map for Pixel Phase 1 in the layouts and fixes the rendering of the same plot


threus:
Thanks,
Tomas for DQM

PR #501, opened 2017-06-19
Fix style of L1TEMU Stage2 Report Summary Map

stahlleiton:
Description:

This PR disables the render plugin of the legacy L1TEMU when the input DQM plot is the L1TEMU Report Summary Map, so that only the L1T Stage2 render plugin is applied on the report summary.


threus:
Thanks,
Tomas for DQM

PR #507, opened 2017-06-28
Render Plugins and Layout for SiStrip APV gain implemented

dimattia:
Greetings,

this PR contains a new render plugin and a new layout for displaying the histograms of the AlcaReco/StripGains DQM directories. It concerns the inclusion of two new files under dqmgui/style and dqmgui/layouts and the modification 

   server-conf-dev.py
   server-conf-online.py

to include the new layout.


threus:
hi @dimattia, since you want the APV layout only in offline, if you rename the layout file to sistripAPVgain_T0_layouts.py, you can drop the changes in both server config files, as it will be taken automatically by this line:
LAYOUTS += glob("%s/layouts/*_T0_layouts.py" % CONFIGDIR)

mmusich:
hi @threus for my own education

> since you want the APV layout only in offline

we want it also in the development GUI (https://cmsweb.cern.ch/dqm/dev/), since that is where the Multi-Run harvesting uploads the data. Does your comment above still holds?

dimattia:
Hi @threus 

I did implemented your suggestion.

dimattia:
Hi @mmusich 

I did assumed so because the DQM GUI from MRH shows the ALCARECO histograms out of the Offline workspace. Anyway let's wait for Thomas to comment.

threus:
great, thanks @dimattia !
@mmusich yep, my comment does hold - looking at the dev server config file, it loads the _T0_ layouts as well, see:
https://github.com/dmwm/deployment/blob/master/dqmgui/server-conf-dev.py#L9
I was not specific, apologies - by the "offline" I meant everything except the online DQM...

mmusich:
> looking at the dev server config file, it loads the T0 layouts as well

wonderful. Thanks for the clarification.

vanbesien:
Hi Marco and Marco,

The dev GUI is the development GUI and is considered a testing GUI.
For the first tests of the multirun harvests, the data was obviously not sent to the real production, but to the test system instead.

If this is becoming a real production workflow, we should move it to the production GUI.
Meaning: Layouts should be added/updated for the offline GUI, not for the dev GUI.
The real production data should be uploaded to the offline GUI.

@rovere @cerminar

Thanks,

Broen

mmusich:
@vanbesien, based on the discussion above I think this:

> Layouts should be added/updated for the offline GUI, 

is already the case.

As for 

>  The real production data should be uploaded to the offline GUI.

that's in the hands of Gianluca / AlCa/DB

vanbesien:
@mmusich I agree on both your points.

Cheers, B

cerminar:
Hello @mmusich @vanbesien 
I can certainly move to the production GUI, no problem. I guess it should be a transparent change... just need to change the target from https://cmsweb.cern.ch/dqm/dev/ to https://cmsweb.cern.ch/dqm/offline/ is that right?

vanbesien:
Correct

threus:
Thanks,
Tomas for DQM

PR #508, opened 2017-06-28
Adjusting shifter layouts and rendering of few plots

fioriNTU:
This add three new plots to the shifter layouts, and adjust the rendering of few trend plots.

Also an "expert" plot is removed from the shift workspace


threus:
Thanks,
Tomas for DQM

PR #510, opened 2017-07-05
Remove reqmgr2 software update cronjob

amaltaro:
@h4d4 this fixes the problem noticed with that reqmgr2 cronjob. The releases are now updated in a cherrypy thread (as in the previous deployment change), so there is no need to have this cronjob around.

There are no RPMs changes, only this deployment update is needed.


ticoann:
Thanks Alan!

PR #512, opened 2017-07-12
Possible fix to recent issue in Phase1 rendering

fioriNTU:
reintroducing a protection on dynamic_casting. Possibly can fix the issue with the rendering we observed lately


threus:
Thanks,
Tomas for DQM

PR #515, opened 2017-07-21
Fix the ROC 0 grid on BPIX layer 1 minus

chenxvan:
Changed the indication of ROC 0 on BPIX layer I minus side to the right place in the ladder-module coordinate plots in Online GUI. Layer 1 plus and other layers remain the same.


threus:
Thanks,
Tomas for DQM

PR #519, opened 2017-08-03
Fix the ROC 0 location on the single module digi occupancy plot

chenxvan:
Changed the location of the ROC 0 grid on the single module digi occupancy plot on layer 1 minus side


threus:
Thanks,
Tomas for DQM

PR #521, opened 2017-08-05
adding support for new TrackingHLTBeamspotStream for reportSummaryMap, near L28

sikler:
Adding a line in BeamRenderPlugin.cc to handle the new beamhlt client ("TrackingHLTBeamspotStream/"), to treat it in a similar way to the previous BeamMonitor/ and BeamMonitor_PixelLess/ folders.

The motivation was to properly plot reportSummaryMap for TrackingHLTBeamspotStream. It seems that reuse of BeamRenderPlugin.cc with the above addition should solve the problem.

### DIFF
--- a/dqmgui/style/BeamRenderPlugin.cc
+++ b/dqmgui/style/BeamRenderPlugin.cc
@@ -23,8 +23,9 @@ class BeamRenderPlugin : public DQMRenderPlugin {
 
 public:
   virtual bool applies( const VisDQMObject &o, const VisDQMImgInfo & ) {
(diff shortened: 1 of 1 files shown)


vanbesien:
Hi,
Do you have any idea of what this "BeamMonitor_PixelLess/" is?

sikler:
Hi,
I am not sure but based on its naming it should be beamspot monitoring for pixel-less tracking (strip only).
So it is for very special conditions (actually happens from time-to-time at the beginning of the running year.)

vanbesien:
Hi Lina @h4d4 , this can be merged.

Broen

PR #523, opened 2017-08-10
Updated the deploy and manage scripts for CC7

vanbesien:
The manage script basically enables you to start the servers and agents
on the new servers: vocms0738 and vocms0739.
The deploy script was _not_ changed, to the extend that the acron jobs
that do the backups to CASTOR will _not_ be installed on the new servers.
(Even if they were activated, they would not get the correct path on CASTOR)
But of course this will be needed when we do the real migration.
Also some changes to the monitoring scripts. This is basically for the
logging and I just added the new servers where the old servers where.
Also removed some very old servers from those files.

@threus Note: This PR is not needed for the online at all.

### DIFF
--- a/dqmgui/deploy
+++ b/dqmgui/deploy
@@ -16,26 +16,12 @@ deploy_dqmgui_deps()
 deploy_dqmgui_prep()
 {
   extra=
-  case $host in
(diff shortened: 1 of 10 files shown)


threus:
@vanbesien ok noted.

PR #525, opened 2017-08-21
Fixed some potential memory leaks in the DQMInfoRenderPlugin

vanbesien:
Replaced the use of "new" to create Text and Lines by the normal standard cpp way of doing it.


threus:
Thanks,
Tomas for DQM

PR #527, opened 2017-08-23
fixing memory leaks in DT and L1TStage2 renderplugins

threus:
Fix of the memory leaks spotted by @rovere


threus:
Thanks,
Tomas for DQM

PR #532, opened 2017-08-28
fixed possible memory leak in SiPixel renderplugin

threus:
Fixed memory leak found by @rovere


threus:
Thanks,
Tomas for DQM

PR #533, opened 2017-08-30
fix possible memory leak

threus:
fix possible memory leak in Ecal renderplugin


threus:
Thanks,
Tomas for DQM

PR #534, opened 2017-08-30
fix possible memory leak in PixelTiming render plugin

threus:
fix possible memory leak in PixelTiming render plugin

### DIFF
--- a/dqmgui/style/SiPixelPhase1OnlineRenderPlugin.cc
+++ b/dqmgui/style/SiPixelPhase1OnlineRenderPlugin.cc
@@ -122,19 +122,21 @@ public:
         if (n_color == 10) n_color = 15; // skip low saturation stuff in between.
 
         h->SetLineColor(n_color);
-        h->Draw("SAME");
-        // i is bins, so 1 is 1st block = 0 to blocksize-1
-        leg->AddEntry(h,(std::to_string((i-1)*blocksize) + "-" + std::to_string((i*blocksize)-1)).c_str(),"l");
+        h->DrawCopy("SAME");
+        // i is bins, so 1 is 1st block = 0 to blocksize-1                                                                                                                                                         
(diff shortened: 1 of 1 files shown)


dqmgui/style/SiPixelPhase1OnlineRenderPlugin.cc
```diff
@@ -122,19 +122,21 @@ class SiPixelPhase1OnlineRenderPlugin : public DQMRenderPlugin
         if (n_color == 10) n_color = 15; // skip low saturation stuff in between.
 
         h->SetLineColor(n_color);
-        h->Draw("SAME");
-        // i is bins, so 1 is 1st block = 0 to blocksize-1
-        leg->AddEntry(h,(std::to_string((i-1)*blocksize) + "-" + std::to_string((i*blocksize)-1)).c_str(),"l");
+        h->DrawCopy("SAME");
+        // i is bins, so 1 is 1st block = 0 to blocksize-1                                                                                                                                                         
+        leg->AddEntry(h->Clone(),(std::to_string((i-1)*blocksize) + "-" + std::to_string((i*blocksize)-1)).c_str(),"l");
```
rovere: No, how many copies are we doing here?

threus: (upper-lower+1), but the Clone() should be owned by the canvas/pad/legend and deleted when that owner is destroyed. At least I understood that from ROOT docu/fora, but it's not clear. That's why I wanted to use Broens' stat machinery. If you already know it's a fail, then no testing is needed and I will think of something different..

vanbesien:
+1
Effectively fixes the leak.

vanbesien:
Also did successful 1-on-1 comparison of an output png before/after for one of the plots for a recent run.

threus:
Thanks,
Tomas for DQM

PR #535, opened 2017-08-31
L1T render plugin upgrade

thomreis:
L1T muon render plugin upgrade to set the texthist option for error summaries.
Also sets the y axis to log scale for object comparison pT histograms.


threus:
Thanks,
Tomas for DQM

PR #540, opened 2017-09-19
Update devvm deployment

juztas:
1) First error is with EMI repo: http://emisoft.web.cern.ch/emisoft/dist/EMI/3/sl6/x86_64/base/repodata/repomd.xml: [Errno 14] PYCURL ERROR 22 - The requested URL returned error: 404 Not Found
       a) EMI grid middleware is taken care by other repos, which are by default on any cern slc6 based vm: epel, slc6-os, EGI-trustanchors.
       b) Removing also slc5 (no point of keeping it)
       c) Not sure about SLC7, but in any case there was no another repo for it.
2) Already for some time if you want to request host certificate automatically using openssl, there was a change needed in openssl.cnf. Append these configuration lines and also use that file in openssl command.

After these changes, it succeeded to deploy devvm without any issues on a new booted VM. Also, could you please update configuration: https://cms-http-group.web.cern.ch/cms-http-group/tutorials/environ/vm-setup.html ? It seems a bit outdated

### DIFF
--- a/system/deploy
+++ b/system/deploy
@@ -236,7 +236,21 @@ system_install_hostcert()
          cd $tmp
          note "WARNING: requesting now a new grid host certificate via gridca.cern.ch"
          note "WARNING: it will use your kerberos credentials for the Single Sign On."
-         openssl req -new -subj $cn -out $req -nodes -sha512 -newkey rsa:2048 -keyout hostkey.pem
+         # 2017.09.19 To be able to use automatic host cer request, openssl config has to be modified.
+         cp /etc/pki/tls/openssl.cnf /tmp/openssl.cnf
+         echo "
+[req]
+req_extensions = v3_req
+
+[ v3_req ]
+basicConstraints = CA:FALSE
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment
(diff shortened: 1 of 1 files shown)


amaltaro:
This doesn't work:
```
++ openssl req -new -subj /CN=alan-cloud2.cern.ch -out alan-cloud2.csr -nodes -sha512 -newkey rsa:2048 -keyout hostkey.pem -config /tmp/openssl.cnf
Error Loading request extension section v3_req
140237806491464:error:22097082:X509 V3 routines:DO_EXT_NCONF:unknown extension name:v3_conf.c:124:
140237806491464:error:22098080:X509 V3 routines:X509V3_EXT_nconf:error in extension:v3_conf.c:93:name=isubjectAltName, value=@alt_names
+++ err=1
+++ note 'ERROR: installation failed with exit code 1.'
```

but after it does after merging multiple sections into only 1.

If Justas doesn't fix this problem, I might get back to it with a patch in a few days. I have some new nodes to commission.

amaltaro:
@h4d4 I think we can just close this one, since the fix was mergeed in another PR.
Thanks to Justas btw, the other patch was almost the same ;)

PR #541, opened 2017-09-20
Replace --exec by --execute in wmagent manage script

amaltaro:
Seeing these kind of warnings when deploying an agent at FNAL CC7 nodes:
```
Installing the mysql schema...
Socket file exists, proceeding with schema install...
Info: Using unique option prefix 'exec' is error-prone and can break in the future. Please use the full name 'execute' instead.
Info: Using unique option prefix 'exec' is error-prone and can break in the future. Please use the full name 'execute' instead.
Installing WMAgent Database: wmagent
```

this patch should fix this issue.
@h4d4 don't merge it yet please.
@ticoann FYI


amaltaro:
Seems to work just fine. Seangchan, please give your final blessing.

ticoann:
Patch looks fine to me. (Just wondering whether this was causing problem in slc7?) Please merge.

amaltaro:
It doesn't cause problems, yet :)
@h4d4 please merge

PR #558, opened 2017-11-03
Workaround for unique id clash

amaltaro:
Lina, as we discussed, this change is needed to avoid a unique ID clash when adding a few users (including reqmgr2ms one).

The problem is kind of a mistery, but as Todor explained, even when we add a linux user locally only, the machine looks up for that ID in the LDAP database and might complain about unique ID constraint.


amaltaro:
FYI @ticoann @vkuznet

PR #565, opened 2017-11-26
Add unified config and changed name of the log

vkuznet:


amaltaro:
Valentin, please add @ticoann  and I to these deployment changes, such that we can be in the loop of what is changing and what we have to ask Lina to push to testbed. Thanks!

PR #570, opened 2017-12-13
remove two plots from SiStrip offline quickcollection

skhalil:
Fix/remove two plots from SiStrip offline quick collection.


threus:
Thanks,
Tomas for DQM

PR #572, opened 2017-12-20
change manage and config script

ticoann:
@vkuznet, please review.


ticoann:
@vkuznet, Hi Valentin,

Where is this file(unifiedConfiguration.json)?

https://github.com/ticoann/deployment/blob/dbd048e1d4ae641a73a660818bb6ddbdd81058c8/reqmgr2ms/manage#L49

I don't see in the repository.
https://github.com/dmwm/deployment/tree/master/reqmgr2ms

ticoann:
I used https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/config.json
but isn't that supposed to be in the deployment directory?

ticoann:
Anyway, I think I will move that to AUX db as well if that is OK.

vkuznet:
Seangchan,
I think I never put it into deployment/reqmgr2ms. But we do have "default"
config in
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/MicroService/Unified
see
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/Common.py#L37
The file is
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/config.json
I think we should move it to deployment/reqmgr2ms though.
Best,
Valentin.

PR #585, opened 2018-02-22
New option for reqmgr2 manage script; push default docs to reqmgrAux

amaltaro:
Drawback is that every single backend will push these documents, though it's not a big deal...
I still have to test it.

### DIFF
--- a/reqmgr2/deploy
+++ b/reqmgr2/deploy
@@ -1,6 +1,13 @@
 # vim: set ft=sh sw=2 ts=8 et :
 deploy_reqmgr2_variants="default prod preprod dev"
 
+case $variant in
+  prod )    base_url="https://cmsweb.cern.ch"         dbs_ins="prod";;
+  preprod ) base_url="https://cmsweb-testbed.cern.ch" dbs_ins="int";;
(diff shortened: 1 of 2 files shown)


amaltaro:
@ticoann can you please review these changes?
I did some tests and I see the correct command in the post log now, though it fails because the WMCore PR has to be merged first and we'd need to cut a new tag/rpm.

amaltaro:
@h4d4 don't merge this yet. You can proceed with the testbed deployment without this patch.
I still have to figure out why I get this complaint
```
Traceback (most recent call last):
  File "/data/srv/current/apps/reqmgr2/bin/reqmgr-put-default-config", line 29, in <module>
    key_file=os.getenv('X509_USER_KEY'))
  File "/data/srv/HG1802d-comp8/sw.amaltaro/slc7_amd64_gcc630/external/python/2.7.13/lib/python2.7/httplib.py", line 1249, in __init__
    context.load_cert_chain(cert_file, key_file)
ssl.SSLError: [SSL] PEM lib (_ssl.c:2709)
```

when running the wmcore script in the deployment. While it works fine in a standalone way.

amaltaro:
This is not going to work, so I'm closing it.

PR #599, opened 2018-04-09
Adding calo object histograms to L1T QuickCollection

vukasinmilosevic:
Additional calorimeter objects have been added to the main summary plots of the L1T online DQM.
A plot for:
L1EG pT, phi-vs-eta, BX
L1Jets pT, phi-vs-eta, BX
L1Tau pT, phi-vs-eta, BX
L1ETM pT, phi-vs-eta, BX
L1HTT pT, phi-vs-eta, BX
L1ETMHF pT, phi-vs-eta, BX

PR #600, opened 2018-04-10
Adding calo object DE agreement summary plots to the L1TEMU QuickCollection/Shifter

vukasinmilosevic:
Adding calo object data to emulator agreement summary plots to the L1TEMU QuickCollection and the shifter collection.
Also, adding the Calo Layer 2 prefix in histogram names for the L1T QuickCollection.
Requires PR #599


vukasinmilosevic:
Hi @threus, should I close this one and open a new one? The conflict came from something that was not affected by #599 (or maybe I'm missing something).

threus:
I believe the conflict came from using the same commits as in #599. Seems like the only relevant commits here are

vukasinmilosevic:
Hi @threus, yes they are. I will close this one and open a new one

threus:
wait - I have to retract my last comment. I assumed that #599 is already merged, which is not true. Apologies for confusion. Then the conflict is because of something else.

vukasinmilosevic:
No problem. I have an additional message that states that the conflict comes from:
dqmgui/layouts/shift_l1temulator_layout.py

PR #602, opened 2018-04-17
Load reqmgr2 secrets file for AMQ settings

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/8569
With this change, reqmgr2 should be able to properly create the secrets file and use it in the service. The file format is the same as those 3 echo lines.
@h4d4 I still have to provide you two files, one for testbed and one for production (I understand you need to write them to a specific cmsweb area that will be accessed during the deployment).

I tested it in my VM, with the secrets under /data/auth/reqmgr2/ and it works fine.
@h4d4 @ticoann can you please review?

### DIFF
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -16,15 +16,12 @@ DBS_INS = "@@DBS_INS@@"
 COUCH_URL = "%s/couchdb" % BASE_URL
 LOG_DB_URL = "%s/wmstats_logdb" % COUCH_URL
 LOG_REPORTER = "reqmgr2"
-USER_AMQ = "@@USER_AMQ@@"
-PASS_AMQ = "@@PASS_AMQ@@"
-AMQ_TOPIC = "@@TOPIC@@"
 AMQ_HOST_PORT = [('dashb-mb.cern.ch', 61113)]
 
 ROOTDIR = __file__.rsplit('/', 3)[0]
 # load AMQ credentials
 sys.path.append(path.join(ROOTDIR, 'auth/reqmgr2'))
-from ReqMgr2Secrets import *
+from ReqMgr2Secrets import USER_AMQ, PASS_AMQ, AMQ_TOPIC
 
 config = Configuration()
 
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -6,21 +6,26 @@ Everything configurable in ReqMgr is defined here.
 
 import socket
 import time
+import sys
 from WMCore.Configuration import Configuration
 from os import path
 
 HOST = socket.gethostname().lower()
 BASE_URL = "@@BASE_URL@@"
 DBS_INS = "@@DBS_INS@@"
-USER_AMQ = "@@USER_AMQ@@"
-PASS_AMQ = "@@PASS_AMQ@@"
 COUCH_URL = "%s/couchdb" % BASE_URL
 LOG_DB_URL = "%s/wmstats_logdb" % COUCH_URL
 LOG_REPORTER = "reqmgr2"
-AMQ_TOPIC = '/topic/cms.jobmon.wmagent'
+USER_AMQ = "@@USER_AMQ@@"
+PASS_AMQ = "@@PASS_AMQ@@"
+AMQ_TOPIC = "@@TOPIC@@"
 AMQ_HOST_PORT = [('dashb-mb.cern.ch', 61113)]
 
 ROOTDIR = __file__.rsplit('/', 3)[0]
+# load AMQ credentials
+sys.path.append(path.join(ROOTDIR, 'auth/reqmgr2'))
+from ReqMgr2Secrets import *
+
 config = Configuration()
 
 main = config.section_("main")
@@ -178,7 +183,10 @@ if HOST.startswith("vocms0136") or HOST.startswith("vocms0731") or HOST.startswi
     heartbeatMonitor.central_logdb_url = LOG_DB_URL
     heartbeatMonitor.log_reporter = LOG_REPORTER
     # AMQ MonIT settings
-    heartbeatMonitor.post_to_amq = True if HOST.startswith("vocms0136") else False  # only production nodes!
+    if HOST.startswith("vocms0731") or HOST.startswith("vocms0136"):
+        heartbeatMonitor.post_to_amq = True
+    else:
+        heartbeatMonitor.post_to_amq = False
     heartbeatMonitor.user_amq = USER_AMQ
     heartbeatMonitor.pass_amq = PASS_AMQ
     heartbeatMonitor.topic_amq = AMQ_TOPIC
--- a/reqmgr2/deploy
+++ b/reqmgr2/deploy
@@ -17,10 +17,6 @@ deploy_reqmgr2_prep()
 
 deploy_reqmgr2_sw()
 {
-  deploy_pkg \
-    -a dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem \
-    -a dmwm-service-key.pem:wmcore/dmwm-service-key.pem \
-    comp cms+reqmgr2
 
   if grep -rq "replace me" $project_auth; then
     note "WARNING: replace certificates in $project_auth with real ones"
@@ -32,6 +28,13 @@ deploy_reqmgr2_sw()
     dev )     base_url="https://cmsweb-dev.cern.ch"     dbs_ins="dev";;
     * )       base_url="https://`hostname -f`"          dbs_ins="private_vm";;
   esac
+
+  deploy_pkg \
+    -a dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem \
+    -a dmwm-service-key.pem:wmcore/dmwm-service-key.pem \
+    -a ReqMgr2Secrets.py:reqmgr2/ReqMgr2Secrets.py \
+    comp cms+reqmgr2
+
   perl -p -i -e "s{\"\@\@BASE_URL\@\@\"}{\"$base_url\"}g; \
                  s{\"\@\@DBS_INS\@\@\"}{\"$dbs_ins\"}g"   \
        $root/$cfgversion/config/$project/config.py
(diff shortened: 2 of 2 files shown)


reqmgr2/deploy
```diff
@@ -32,6 +28,13 @@ deploy_reqmgr2_sw()
     dev )     base_url="https://cmsweb-dev.cern.ch"     dbs_ins="dev";;
     * )       base_url="https://`hostname -f`"          dbs_ins="private_vm";;
   esac
+
+  deploy_pkg \
+    -a dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem \
+    -a dmwm-service-key.pem:wmcore/dmwm-service-key.pem \
+    -a ReqMgr2Secrets.py:reqmgr2/ReqMgr2Secrets.py \
```
ticoann: Alan, Lina, what does this part do?

ticoann:
Sorry Alan that you had to do this. I started yesterday but got distracted with something else. :-)
I think my approach was little different then this (not better). 

FYI.
https://github.com/dmwm/deployment/compare/master...ticoann:amq_config?expand=1

amaltaro:
No problem. I think your approach would work too :-)

reqmgr2/deploy
amaltaro: I can't even find the deploy_pkg script, but I think this is where the secrets is copied from /data/srv/auth to the current deployment (/data/srv/current/auth), and permissions/ownership are set. Anyways, it's better if Lina can confirm.

reqmgr2/config.py
```diff
@@ -6,21 +6,26 @@
...
-AMQ_TOPIC = '/topic/cms.jobmon.wmagent'
+USER_AMQ = "@@USER_AMQ@@"
+PASS_AMQ = "@@PASS_AMQ@@"
+AMQ_TOPIC = "@@TOPIC@@"
 AMQ_HOST_PORT = [('dashb-mb.cern.ch', 61113)]
 
 ROOTDIR = __file__.rsplit('/', 3)[0]
+# load AMQ credentials
+sys.path.append(path.join(ROOTDIR, 'auth/reqmgr2'))
+from ReqMgr2Secrets import *
```
ticoann: Wouldn't this fail if we don have secrets in vm? Or will we always have secrete files?
In that case, we have to put different contents for different instance isn't it. (For vm test, empty?)

reqmgr2/config.py
```diff
@@ -178,7 +183,10 @@
     heartbeatMonitor.central_logdb_url = LOG_DB_URL
     heartbeatMonitor.log_reporter = LOG_REPORTER
     # AMQ MonIT settings
-    heartbeatMonitor.post_to_amq = True if HOST.startswith("vocms0136") else False  # only production nodes!
+    if HOST.startswith("vocms0731") or HOST.startswith("vocms0136"):
+        heartbeatMonitor.post_to_amq = True
+    else:
+        heartbeatMonitor.post_to_amq = False
```
ticoann: Maybe we can use this setting to have default values for USER and PASS ("" each?) - Although it wouldn't matter since it won' try to instantiate StormAMQ, right?

reqmgr2/config.py
```diff
@@ -178,7 +183,10 @@
...
     heartbeatMonitor.user_amq = USER_AMQ
     heartbeatMonitor.pass_amq = PASS_AMQ
```
ticoann: So are you saying content of ReqMgr2Secrets.py would be global variable.

```
USER_AMQ = blah
PASS_AMQ = blah
```

reqmgr2/config.py
```diff
@@ -6,21 +6,26 @@
...
 HOST = socket.gethostname().lower()
 BASE_URL = "@@BASE_URL@@"
 DBS_INS = "@@DBS_INS@@"
-USER_AMQ = "@@USER_AMQ@@"
-PASS_AMQ = "@@PASS_AMQ@@"
 COUCH_URL = "%s/couchdb" % BASE_URL
 LOG_DB_URL = "%s/wmstats_logdb" % COUCH_URL
 LOG_REPORTER = "reqmgr2"
-AMQ_TOPIC = '/topic/cms.jobmon.wmagent'
+USER_AMQ = "@@USER_AMQ@@"
```
ticoann: This is not needed, isn't it. Since ReqMgr2Secrets.py content will overwrite this anyway

reqmgr2/config.py
```diff
@@ -6,21 +6,26 @@
...
+PASS_AMQ = "@@PASS_AMQ@@"
```
ticoann: Same as above

reqmgr2/deploy
```diff
@@ -65,7 +68,14 @@ deploy_reqmgr2_post()
...
-    */*-cert.pem ) echo "replace me with your dmwm service certificate" ;;
-    */*-key.pem )  echo "replace me with your dmwm service key" ;;
+    */*-cert.pem )
+      echo "replace me with your dmwm service certificate" ;;
+    */*-key.pem )
+      echo "replace me with your dmwm service key" ;;
+    */ReqMgr2Secrets*.py )
+      echo 'USER_AMQ = "FOO_FIXME"'
+      echo 'PASS_AMQ = "BAR_FIXME"'
+      echo 'AMQ_TOPIC = "FOOBAR_FIXME"'
```
ticoann: What is the purpose of doing the echo? Just for information?

ticoann:
Alan, sorry for a lot of comments. 😄 Mostly due to my lack of understanding.

reqmgr2/config.py
amaltaro: I think the idea is to just keep people aware that these arguments should exist (and to avoid pylint complaints lower in the config file).

reqmgr2/config.py
amaltaro: and if you prefer, we can either remove or set it to something like FIXME, REPLACE-ME etc

reqmgr2/config.py
```diff
@@ -6,21 +6,26 @@
...
+AMQ_TOPIC = "@@TOPIC@@"
```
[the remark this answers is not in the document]
amaltaro: Oh it does change between cmsweb instances. Check out my private email and you'll see where it's set ;)

reqmgr2/config.py
amaltaro: Good question, let me test it without a secret file in my VM.

reqmgr2/config.py
ticoann: I think it is more confusing since it is replaced right away.
It would be better import specific things below. (pylint check will complain anyway)

```
from ReqMgr2Secrets import *
-->
from ReqMgr2Secrets import USER_AMQ, PASS_AMQ, 

```

reqmgr2/config.py
amaltaro: Exactly, we don't instantiate StompAMQ, see: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py#L106

I think I didn't get your point about USER and PASS. They already have a default value from the top of this config

reqmgr2/config.py
amaltaro: Yes, I tested these variables at this level and I checked the values come from the secrets file, not the default one set at the top of this config. (Note, code stolen from DBS setup :P)

reqmgr2/deploy
amaltaro: Hmm, this might answer your question above. I think it will create a file with this content in case there is no secret in the backend. Let me check and get back to you.

reqmgr2/config.py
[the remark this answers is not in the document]
ticoann: Hmm, I didn't see in your email, (sent to Lina and me). Anyway, if this needs to be changes for different instances,
deploy script also needs to be changed. Line 38 and above.

```

perl -p -i -e "s{\"\@\@BASE_URL\@\@\"}{\"$base_url\"}g; \
                  s{\"\@\@DBS_INS\@\@\"}{\"$dbs_ins\"}g" 
...

```

reqmgr2/deploy
ticoann: Hmm, I am not sure it will right the content in the file.  Anyway, We shouldn't create the content here, since the whole point is hiding secretes.

reqmgr2/config.py
amaltaro: Sounds good, I'll update it.

reqmgr2/config.py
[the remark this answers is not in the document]
amaltaro: I've put it in the secrets file too. That's why it's not replaced in this perl line.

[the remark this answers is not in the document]
ticoann: Hmm, I don't quite get it, Are there 2 different secrete files for each instance then?

ticoann:
I still don't understand deploy script part but I guess we should wait for the Lina's comments. Other than that, patch looks good, Alan

amaltaro:
It took me basically the whole day to make it working, and I still don't quite understand that :-)

reqmgr2/deploy
h4d4: @ticoann @amaltaro 
Alan, accord to deployment logs you're right. 
Look for example[1] this is part of the log execution of deploy_reqmgr2_sw for the last tag deployed on testbed.
Basically, during this step the secrets files are installed in to the current tag.

[1] +++ deploy_reqmgr2_sw
+++ deploy_pkg -a dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem -a dmwm-service-key.pem:wmcore/dmwm-service-key.pem comp cms+reqmgr2
+++ local authfiles=
+++ local authfile=
+++ local link=
+++ '[' 6 -ge 1 ']'
+++ case $1 in
+++ authfiles=' dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem'
+++ shift
+++ shift
+++ '[' 4 -ge 1 ']'
+++ case $1 in
+++ authfiles=' dmwm-service-cert.pem:wmcore/dmwm-service-cert.pem dmwm-service-key.pem:wmcore/dmwm-service-key.pem'
+++ shift
+++ shift
+++ '[' 2 -ge 1 ']'
+++ case $1 in
+++ break
+++ '[' 2 = 0 ']'
++++ echo comp
++++ sed -e 's/comp/comp/; n'
+++ local repo=comp
++++ echo comp
++++ sed 's/^comp//; s/^\([^.]\)/.\1/; s/^/sw/; n'
+++ local label=sw
+++ local swarea=/data/srv/beHG1805a/sw
+++ local ver=auto
+++ local pkg=cms+reqmgr2
+++ case $ver:$pkg in
+++ case $ver:$release in
+++ set -e
+++ inrepo comp /data/srv/beHG1805a/sw
+++ local repo=comp
+++ local swarea=/data/srv/beHG1805a/sw
+++ '[' '!' -f /data/srv/beHG1805a/sw/.bootstrapped ']'
+++ '[' Xauto '!=' Xauto ']'
++++ /data/srv/beHG1805a/sw/common/cmspkg -a slc7_amd64_gcc630 depends cms+comp+HG1805a
++++ fgrep 'Depends: cms+reqmgr2+'
++++ awk -F+ '{print $NF}'

amaltaro:
@ticoann it has been tested and it's good to go.

ticoann:
Thanks Alan, I will add to pull request so Lina can merge it.

PR #606, opened 2018-04-23
Mock workqueue credentials for couchapps push

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/8579
@ticoann can you please review this patch

### DIFF
--- a/workqueue/config.py
+++ b/workqueue/config.py
@@ -3,6 +3,7 @@ _GlobalWorkQueueConfig_
 
 Global cmsweb WorkQueue config.
 """
+from __future__ import print_function
 
 import socket
 import os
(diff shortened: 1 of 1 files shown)


ticoann:
Hmm, why it cannot be import WorkQueueSecrete?

amaltaro:
My best guess is that couchdb user can't access workqueue's credentials (auth stuff), thus it cannot import it.

As I mentioned in the WMCore issue, the PYTHONPATH was properly set by the workqueue/manage script, so there must have been something else between the manage and the wmagent-couchcapp-init scripts that would screw that env var. Or it's the file permissions as my first comment above.

PR #610, opened 2018-04-27
Extend TLine to fix drawing line on plots for Wheel -2 and +2 in RPCRenderPlugin

slezki:
Fixing DrawLine options for Wheel -2 and +2.
@andresib @mileva @jhgoh


slezki:
@threus yes it is final and you can start to test it. it will be used for offline as well.

PR #629, opened 2018-05-23
adding charge per CM

MRD2F:
Adding cluster charge per cm on track in shift strip layouts


threus:
Thanks, as always!
Tomas for DQM

PR #637, opened 2018-06-05
Dissabling das to be instaled on vocsm0740

h4d4:
@vkuznet @amaltaro 
This PR is to disable das/da2sgo on vocms0740, due to in today's production upgrade I found that das/da2sgo was installed on this node. As I understand, it should only run on vomcs074{1,2}

I've disabled this service on vocms0740, in order to avoid alarms I also updated the metrics.

Best Regards, Lina

### DIFF
--- a/das/deploy
+++ b/das/deploy
@@ -21,7 +21,7 @@ deploy_das_sw()
 
 deploy_das_post()
 {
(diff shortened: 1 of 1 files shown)


vkuznet:
Lina, yes this is correct, please go ahead and merge it.

PR #644, opened 2018-06-20
adding plots in the Tracking layout

MRD2F:
Were added two plots in the Tracking layout, indicating the number of tracking hits in the Pixel and Strip separately


threus:
Thanks,
Tomas for DQM

PR #658, opened 2018-08-13
add step chain fix for reqmgr

ticoann:
Tested. @h4d4, Lina, could you check whether I wrote correct production/testbed nodes. 
Each thread only need to run in one vocms box in production/testbed

### DIFF
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -117,9 +117,9 @@ ui_main = ui.section_("main")
 ui_main.application = ui.index
 #ui_main.authz_defaults = {"role": None, "group": None, "site": None, "policy": "dangerously_insecure"}
 
+extentions = config.section_("extensions")
 # Production instance of wmdatamining, must be a production back-end
-if HOST.startswith("vocms0136") or HOST.startswith("vocms0731") or HOST.startswith("vocms0117") or HOST.startswith("vocms0127"):
-    extentions = config.section_("extensions")
+if HOST.startswith("vocms0136") or HOST.startswith("vocms0132") or HOST.startswith("vocms0117") or HOST.startswith("vocms0127"):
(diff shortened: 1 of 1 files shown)


reqmgr2/config.py
```diff
@@ -146,7 +146,18 @@
...
+    parentageFixTask = extentions.section_("parentageFixTask")
+    parentageFixTask.object = "WMCore.ReqMgr.CherryPyThreads.StepChainParentageFixTask.StepChainParentageFixTask"
+    parentageFixTask.dbs_url = data.dbs_url
+    parentageFixTask.reqmgrdb_url = "%s/%s" % (data.couch_host, data.couch_reqmgr_db)
+    parentageFixTask.parentageFixDuration = 60 * 180  # every 3 hours
+    parentageFixTask.log_file = '%s/logs/reqmgr2/parentageFixTask-%s.log' % (__file__.rsplit('/', 4)[0], time.strftime("%Y%m%d"))
+    parentageFixTask.central_logdb_url = LOG_DB_URL
+    parentageFixTask.log_reporter = LOG_REPORTER
+
+if HOST.startswith("vocms0161") or HOST.startswith("vocms0731"):
```
amaltaro: add these `or HOST.startswith("vocms0117") or HOST.startswith("vocms0127"):` here too.

amaltaro:
Lina, with this PR we also have to update this wiki page:

since there is one new thread `ReqMgr2 parentageFixTask` and we also distributed the other threads in different backends.

reqmgr2/config.py
h4d4: @ticoann @amaltaro 
Yes, the nodes production/testbed are well wrote

h4d4:
@amaltaro Yes, I'll update the twiki.

PR #666, opened 2018-09-05
Enable AMQ updates from vocms0161

amaltaro:
Lina, with the recent changes to distribute different threads between different backends, we missed that line that enables updates to the CERN MonIT system. I'm sending you a private email with instruction on what needs to be done.

FYI @ticoann


ticoann:
Sorry Lina, Alan, It was my fault.

PR #673, opened 2018-09-27
Remove acdcserver cleanup cronjob

amaltaro:
Cleanup of ACDC couch docs is taken care by a ReqMgr2 thread (for quite some time already), thus there is no need to keep this - broken - configuration/cronjob around.
@ticoann please review


amaltaro:
Changes tested and it looks good. However, I don't think we need to run another deployment to get this change in. It can wait the cmsweb production upgrade IMO.

ticoann:
Yes that would be fine. Thanks Alan.

amaltaro:
You can either merge it now or after the cmsweb upgrade. Either way is fine for me.

h4d4:
@amaltaro @ticoann 
Thanks for the feedback! merged.

PR #685, opened 2018-10-25
Remove LogDB docs older than 90 days instead of one

amaltaro:
Changes related to https://github.com/dmwm/WMCore/pull/8870 , keeping any logDB documents newer than 90 days. Cleanup `DataKeepDays` that is no longer used.
@ticoann please review


ticoann:
Patch looks good, Thanks Alan.

PR #692, opened 2018-11-14
Write one user entry per line in the authmap.json file

amaltaro:
It looks like the perl module parsing this file and creating a new request header/auth/authz metadata - which is passed to the backends - has a very peculiar way of reading that json file and it requires one entry per line. Since CRIC returns a pretty format, which makes it all the way to authmap.json, some attributes are always missed.

For the record, this is the module parsing the json and creating the necessary information for the backends:
https://github.com/dmwm/deployment/blob/master/frontend/cmsauth.pm

Real change is in the first commit. Second is only aesthetic changes.
Still has to be tested in a clean environemtn.

### DIFF
--- a/frontend/mkauthmap
+++ b/frontend/mkauthmap
@@ -118,13 +118,15 @@ def updateFile(opts, roles, sitemap):
                 if opts.verbose:
                   print "WARNING: role %s is missing from the sitemap" % r
             entry['ROLES'][group] = group_roles
-      roles = json.dumps(struct,indent=2) # converts back to a string
+      # Generate stable JSON representation.
+      enc = json.JSONEncoder(sort_keys=True)
+      jsondata = "[\n " + ",\n ".join(enc.encode(entry) for entry in struct) + "\n]\n"
     try:
       fd, tmpname= mkstemp(dir = os.path.dirname(opts.out))
       tmpfile = os.fdopen(fd, "w")
-      tmpfile.write(roles)
+      tmpfile.write(jsondata)
       tmpfile.close()
-    except(IOError, OSError) as e: 
+    except(IOError, OSError) as e:
       print "An error ocurred:"
       pprint(e)
     try:
--- a/frontend/mkauthmap
+++ b/frontend/mkauthmap
@@ -24,27 +24,27 @@ sitemap = {}
 sites = ''
 
 
-def convertToHeadersStyle(str):
+def convertToHeadersStyle(siteInfo):
     """  
      - all lower case
      - whitespaces and underscores replaced by dashes
     """
-    result = str.replace('_', '-')
-    result = result.replace(' ', '-')
+    result = siteInfo.replace('_', '-').replace(' ', '-')
     return result.lower()
 
 
-def getFecthDetails(file):
-    nline = 0
-    for line in open(file, "r"):
-        nline += 1
-        m = re.match(r"^url (cric?:\S+)$", line)
-    if not m:
-        print "%s: %d: line not understood" % (file, nline)
-        sys.exit(1)
-    ml = m.group(1)
-    url = ml[5:]
-    return url
+def getFecthDetails(fname):
+    with open(fname) as fp:
+        content = fp.readlines()
+
+    for line in content:
+        m = re.match(r"^url (cric?:\S+)$", line.strip())
+        if not m:
+            print "%s: %d: line not understood" % (fname, line)
+            sys.exit(1)
+        ml = m.group(1)
+        url = ml[5:]
+        return url
 
 
 def request(uri):
@@ -142,30 +142,26 @@ def updateFile(opts, roles, sitemap):
         try:
             if os.path.exists(tmpname) and os.path.getsize(tmpname) > 0:  # Ensuring that file exits and is not empty
                 diff = diffFiles(tmpname, opts.out)
-                if (diff == 0):  # fetched content have not changed
-                    try:
-                        os.remove(tmpname)
-                    except(IOError, OSError) as e:
-                        print "An error ocurred:"
-                        pprint(e)
-                elif (diff == -1):  # something wrong was with diff function
-                    try:
-                        os.remove(tmpname)
-                    except(IOError, OSError) as e:
-                        print 'An error ocurred: '
-                        pprint(e)
+                if diff == 0:
+                    # fetched content have not changed
+                    pass
+                elif diff == -1:
+                    # something wrong was with diff function
                     exit(1)
-                else:  # fetched content have changed
-                    try:
-                        command = "cp " + tmpname + " " + opts.out
-                        p = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
-                        stdout = p.communicate()[0]
-                        # print to know if the json have being updated
-                        print "Info: json was updated"
-                        os.remove(tmpname)
-                    except(IOError, OSError) as e:
-                        print 'An error ocurred: '
-                        pprint(e)
+                else:
+                    # fetched content have changed
+                    command = "cp " + tmpname + " " + opts.out
+                    p = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
+                    stdout = p.communicate()[0]
+                    # print to know if the json have being updated
+                    print "Info: json was updated"
+
+                # finally remove the tmp file
+                try:
+                    os.remove(tmpname)
+                except(IOError, OSError) as e:
+                    print 'An error ocurred: '
+                    pprint(e)
         except(IOError, OSError) as e:
             print "An error ocurred"
             pprint(e)
--- a/frontend/mkauthmap
+++ b/frontend/mkauthmap
@@ -3,186 +3,197 @@
 Queries CRIC roles .json and put it into FILE
 """
 
-import os, subprocess, sys
-from traceback import format_exc
-import urllib2, os, os.path, re
 import json
-from urllib2 import Request, urlopen, URLError
+import os
+import os.path
+import re
+import subprocess
+import sys
+import urllib2
 from optparse import OptionParser
+from pprint import pprint  # to debug objects content
 from tempfile import mkstemp
-from shutil import copyfile
+from urllib2 import URLError
 from urlparse import urlsplit, urlunparse
-from pprint import pprint #to debug objects content
 
-#Global vars
+# Global vars
 opts = ''
 args = ''
 roles = ''
 sitemap = {}
 sites = ''
-"""  
- - all lower case
- - whitespaces and underscores replaced by dashes
-"""  
-def convertToHeadersStyle (str):
-  result=str.replace('_','-')
-  result=result.replace(' ','-')
-  return result.lower()
+
+
+def convertToHeadersStyle(str):
+    """  
+     - all lower case
+     - whitespaces and underscores replaced by dashes
+    """
+    result = str.replace('_', '-')
+    result = result.replace(' ', '-')
+    return result.lower()
+
 
 def getFecthDetails(file):
-  nline = 0
-  for line in open(file,"r"):
-    nline += 1
-    m = re.match(r"^url (cric?:\S+)$", line)
-  if not m:
-    print "%s: %d: line not understood" % (file, nline)
-    sys.exit(1)
-  ml  =  m.group(1)
-  url =  ml[5:]
-  return url
+    nline = 0
+    for line in open(file, "r"):
+        nline += 1
+        m = re.match(r"^url (cric?:\S+)$", line)
+    if not m:
+        print "%s: %d: line not understood" % (file, nline)
+        sys.exit(1)
+    ml = m.group(1)
+    url = ml[5:]
+    return url
+
 
 def request(uri):
-  content = ''
-  url = "http:" + uri
-  apimap = {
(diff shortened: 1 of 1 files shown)


amaltaro:
Note this PR was created against the branch with the CRIC/Natalia's changes.

vkuznet:
Alan, could you please post CRIC API which returns "pretty format". I've seen with Rucio server that they not return complaint JSON and I wonder that if the problem with CRIC itself rather then parsing the JSON. I want to see the actual JSON and the API which yields it.

frontend/mkauthmap
```diff
@@ -3,184 +3,196 @@
...
+                                if opts.verbose:
+                                    print "Added role: %s for user %s and group %s " % \
+                                          (sitemap[r], entry['NAME'], group)
+                            else:
+                                if opts.verbose:
+                                    print "WARNING: role %s is missing from the sitemap" % r
+                        entry['ROLES'][group] = group_roles
+            roles = json.dumps(struct)  # converts back to a string
+            # This looks ugly, but it makes one user entry per line in the authmap.json
+            roles = roles.replace('{"DN"', '\n{"DN"')
```
vkuznet: Did you test that you can read back the roles after this transformation? I mean if you write it to file below, can you then read it back as json.load(open(fname))?

amaltaro:
Valentin, this is the API:
http://cms-cric.cern.ch/api/accounts/user/query/?json&preset=roles

and I just checked the json generated and I can json load it without any issues.

I also noticed the previous code had a nice mechanism to add one entry per line, so I changed this patch and now it looks much much better. Still requires one last test though.

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
+    rcode = -1
     try:
-      if os.path.exists(tmpname) and os.path.getsize(tmpname) > 0:#Ensuring that file exits and is not empty
-        diff = diffFiles(tmpname,opts.out)
-        if (diff == 0):#fetched content have not changed
-          try:
-            os.remove(tmpname)
-          except(IOError, OSError) as e:
+        command = "diff " + file1 + " " + file2
+        p = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
```
vkuznet: Are we aware about the consequences here? How often this mkauthmap is use in frontend? We are spawning subprocess and executing diff unix command. If we need to do it in real time I doubt it's going to scale? Would it be better to compare JSONs via function call?

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
+                                group_roles.extend(sitemap[r])
+                                if opts.verbose:
+                                    print "Added role: %s for user %s and group %s " % \
+                                          (sitemap[r], entry['NAME'], group)
+                            else:
+                                if opts.verbose:
+                                    print "WARNING: role %s is missing from the sitemap" % r
+                        entry['ROLES'][group] = group_roles
+            # Generate stable JSON representation.
+            enc = json.JSONEncoder(sort_keys=True)
```
vkuznet: why sorting is required here?

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
+    """
+    if roles:
+        if sitemap:
+            struct = json.loads(roles)
+            for entry in struct:
+                # Only extend the groups interesting to PhEDEx:
+                for group in ['data-manager', 'site-admin', 'phedex-contact']:
+                    group_roles = []
+                    if group in entry['ROLES']:
+                        group_roles.extend(entry['ROLES'][group])
```
vkuznet: extending list is an expensive operations, better is to create a temp dict and assign keys to it instead of storing them as list values. Then, later in a code just use temp_dict.keys() and you'll be done.

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
+                        for r in entry['ROLES'][group]:
+                            if not r.startswith('site:'):
+                                continue
+                            if r in sitemap:
+                                group_roles.extend(sitemap[r])
```
vkuznet: same argument as above about extend vs temp_dict.

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
+        try:
+            if os.path.exists(tmpname) and os.path.getsize(tmpname) > 0:  # Ensuring that file exits and is not empty
+                diff = diffFiles(tmpname, opts.out)
+                if (diff == 0):  # fetched content have not changed
+                    try:
+                        os.remove(tmpname)
+                    except(IOError, OSError) as e:
+                        print "An error ocurred:"
+                        pprint(e)
+                elif (diff == -1):  # something wrong was with diff function
```
vkuznet: this block is identical to previous one, therefore they can be combined, e.g.
```
if diff == 0 or diff == -1:
```

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
-    nline += 1
-    m = re.match(r"^url (cric?:\S+)$", line)
-  if not m:
-    print "%s: %d: line not understood" % (file, nline)
-    sys.exit(1)
-  ml  =  m.group(1)
-  url =  ml[5:]
-  return url
+    nline = 0
+    for line in open(file, "r"):
```
vkuznet: this open call is never closed therefore should be refactored as
```
with open(file, 'r') as istream:
   for line in istream:
       # the rest goes here
```

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
-  return result.lower()
+
+
+def convertToHeadersStyle(str):
+    """  
+     - all lower case
+     - whitespaces and underscores replaced by dashes
+    """
+    result = str.replace('_', '-')
+    result = result.replace(' ', '-')
```
vkuznet: can be combined into single statement, e.g.
```
result = str.replace('_', '-').replace(' ', '-')
```

frontend/mkauthmap
```diff
@@ -3,184 +3,197 @@
...
- - all lower case
- - whitespaces and underscores replaced by dashes
-"""  
-def convertToHeadersStyle (str):
-  result=str.replace('_','-')
-  result=result.replace(' ','-')
-  return result.lower()
+
+
+def convertToHeadersStyle(str):
```
vkuznet: bad usage of reserved python keyword `str`. It defines the generic string type and SHOULD NOT be used as argument name.

frontend/mkauthmap
vkuznet: bad usage of reserved python keyword `file`. It should not be used as an argument of a function.

frontend/mkauthmap
amaltaro: That's a valid concern Valentin and we do NOT have such code in master. 
It looks like this has been added by Lina.
About how often it runs, in my VM it's set to run every 4min, I guess the same apply to CMSWEB (from each frontend)

frontend/mkauthmap
amaltaro: I assume for the comparison.

frontend/mkauthmap
amaltaro: I think it happens for less than 5% of the CMS user, so it shouldn't be a big deal. Anyways, this was Natalias code as well...

frontend/mkauthmap
amaltaro: Yep

amaltaro:
Thanks for the review, @vkuznet . However, all the bad code you pointed out is something that wasn't added by me in this PR.

My real changes are in the first commit only. Second commit is just an automatic code refactoring that I trigger in my IDE to make the code better, of course it only applies conventions, not logic.

Anyways, I'll update some of your comments tomorrow and that's it. I won't spend any more time on it.
If you want to make it better, feel free to take these changes and work in your branch

amaltaro:
@h4d4 @vkuznet @nataliaratnikova and @kdziedzi
Summary of the commits as follows:
1) real change/fix to write out one user entry per line in the json file
2) major code reformat (no logic changes, just python conventions). Don't waste time looking at this one!
3) applied some better practice coding changes as suggested by Valentin (a few logic changes in here)
4) got rid off the `diff` subprocess, meaning you no longer can see what are the differences between runs of that cronjob

Some of these commits bring back some logic that we had (still in master) previously in the script.
If you still want to know the exact diff, I can remove the 4th commit. However, as Valentin said, it might be dangerous to keep that code as is.

I just tested it in my VM and it seems to be working Ok (besides the `-q` option in the cron entry that keeps appearing in my dev VM!).
I'm done with changes to this, please review and merge at your convenience.

h4d4:
@amaltaro @vkuznet Hello Alan, Valentin.

Thank you for the reviews, I really appreciate that!
I'll follow recommendations with aims to get a better code.
And proceed with the corresponding tests.

Best Regards, Lina.

frontend/mkauthmap
```diff
@@ -3,184 +3,173 @@
...
+    result = siteInfo.replace('_', '-').replace(' ', '-')
+    return result.lower()
+
+
+def getFecthDetails(fname):
+    with open(fname) as fp:
+        content = fp.readlines()
+
+    for line in content:
+        m = re.match(r"^url (cric?:\S+)$", line.strip())
```
vkuznet: this is waste of time since every time you call the function and for every content you generate regexp. Instead compile regexp once and use it, e.g.

```
# in a global scope, at the begining of the file
PAT_URL = re.compile(r"^url (cric?:\S+)$")
...
def getFetchDetails(fname):
     with open(fname) as fp:
         while True:
               line = fp.readline()
               if not line: break
               m = PAT_URL.match(line.strip())
```

In my code I also get rid of content since you load all lines from file at once and it means you allocate the memory. Instead I shown that you need to "scan" file line by line without memory allocation for entire content.

Then, just adjust the code further as you already have in body of this function.

frontend/mkauthmap
```diff
@@ -3,184 +3,173 @@
...
+    os.umask(val)
+    return val
+
+
+def convertToHeadersStyle(siteInfo):
+    """  
+     - all lower case
+     - whitespaces and underscores replaced by dashes
+    """
+    result = siteInfo.replace('_', '-').replace(' ', '-')
```
vkuznet: Make one line return
```
return siteInfo.replace('_', '-').replace(' ', '-').lower()
```

frontend/mkauthmap
vkuznet: Alan, I understand the value of your time, therefore let Natalia to adjust it. But we're talking about very sensitive subject here, the frontend, every bit matter, otherwise we will continue struggle with interminent interruptions.

frontend/mkauthmap
vkuznet: so, if we get rid of `diff`, then I would assume we don't need this sorting. Let's remove it.

frontend/mkauthmap
amaltaro: That regex is evaluated against a single line file. Thus the gain in this case is close to 0. I suggest we forget about it.

frontend/mkauthmap
amaltaro: That's a matter of taste ;) And it's just a small string, so let's forget this one too.

amaltaro:
Valentin, this script runs in 2 or 3 seconds, so I don't think optimizations that will buy us half a second make much difference here.
In addition to that, we don't plan to use it for too long, so I suggest we just go with this functional PR as it is. It's anyways much much better than the previous version.

vkuznet:
Alan, everyone should strive to produce a high quality code regardless of its usage. If a person will look for excuses, neither person or code will ever improve. Even we'll use it for short period of time, people may copy-and-paste its parts later and we again will have bad code around. I know that you're quite tough in other PRs, so be yourself and don't let the circumstances to lower the code quality.

kdziedzi:
Valentin,
Honestly, let's leave it for the moment.
The code is only there for Phedex and I hope it will disappear with the authmap rewrite.
Alan has basically taken up a task that was abandoned due to miscommunication and fixed it, so I'd cut him some slack.
________________________________

frontend/mkauthmap
nataliaratnikova: Hi All.
I see that Alan already converted json into one-line format both Lina's original code and my changes on top of it, which were following Lina's format. I did a quick test and it is good enough the purpose of my patch: compare CRIC/SiteDB API contents coming from via the headers and from the local API dumps. 
I tested Alan's latest version: commit 92ca5cc3b53ded57e7c200e06120ea7b4554e91f (HEAD -> entry-line, origin/entry-line)
it works both for old and new PhEDEx secmod. If there is anything else that needs to be fixed or testes I can look into this closer to the point when this is goign to be deployed. Currently my main focus is on testing PhEDEx new SecMod finally deployed today.

kdziedzi:
Natalia, it will be redeployed according to the previous schedule on the 26th.
Sorry it was deployed earlier  - there was a confusion with the PRs.
Kate

________________________________

h4d4:
@kdziedzi @nataliaratnikova 
Hello Kate, 

then to summarize are we agree to merged, the commits [1]?

Best Regards, Lina.
https://github.com/dmwm/deployment/pull/692/commits

kdziedzi:
Please go ahead.
Cheers,
Kate

________________________________

h4d4:
@kdziedzi @nataliaratnikova 
Thanks!

PR #701, opened 2018-11-30
Hg1811 sitedb to cric

h4d4:


h4d4:
This PR is to comment prints where the file is updated, there is something that need to be review due to 
messages are displayed even if the file is not updated

PR #706, opened 2018-12-14
Add support for x509-scitoken-issuer

vkuznet:


frontend/deploy
```diff
@@ -45,30 +45,40 @@ deploy_frontend_post()
...
   $nogroups || { sudo chown -R $USER:_config $PWD/{*.conf,htdocs,etc} || true; }
 
+  # fix path in x509_scitokens_issuer configs before we'll make server configuration
+  perl -p -i -e "s|\@STATEDIR\@|$project_state|g" \
+    $project_config/x509_scitokens_issuer/x509_scitokens_issuer_app.json
+  perl -p -i -e "s|\@STATEDIR\@|$project_state|g" \
+    $project_config/x509_scitokens_issuer/x509_scitokens_issuer_http.conf
+  perl -p -i -e "s|cmsweb.cern.ch|$issuer|g" \
+      $project_config/x509_scitokens_issuer/openid-configuration
+  perl -p -i -e "s|cmsweb.cern.ch|$issuer|g" \
```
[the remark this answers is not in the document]
vkuznet: This is done, see https://github.com/dmwm/deployment/pull/706/files#diff-6ff35e31f83270358c5469df6c997b0bR54

frontend/deploy
```diff
@@ -86,14 +96,45 @@ deploy_frontend_post()
...
+
+  mkdir -p $project_state/wsgi
+  mkdir -p $project_state/logs
+  mkdir -p $project_state/x509_scitokens_issuer
+  mkdir -p $project_state/htdocs/.well-known
+  mkdir -p $project_state/htdocs/oauth2
+
+  scitokens-admin-create-key --create-keys --pem-private --pem-public > /tmp/issuer.pem
+  scitokens-admin-create-key --private-keyfile /tmp/issuer.pem --jwks-private > /tmp/issuer_key.jwks
+  scitokens-admin-create-key --public-keyfile /tmp/issuer.pem --jwks-public > /tmp/issuer_public.jwks
```
bbockelm: This is probably fine for the testbed setups, but may not work in production.

The issuer key is a public / private keypair used to verify the token validity by remote endpoints.  Typically the public key portion is downloaded from the issuer once every few hours.

So, if we clobber the production key every time we redeploy a frontend, it will start issuing tokens with a public key that is unknown to the remote endpoints until a cache refresh occurs.

Luckily, we can handle this with a key rotation strategy - you add the new public key to the list of public keys and then, a few hours later, it can be used for signing.

vkuznet: In this case I think we should discuss with Lina the strategy. We do keep our secret stuff separately, e.g. DBParam files, and we can generate key once and then re-use them. If that what you want I would suggest that we deploy in this way and let Lina tweak further the changes to adopt copying keys from a safe place.

frontend/deploy
```diff
@@ -86,14 +96,45 @@ deploy_frontend_post()
...
+
+  cp -f /tmp/issuer_public.jwks $project_state/htdocs/oauth2/certs
```
bbockelm: Note that this file can be used for multiple keys.  Example:
```
{
    "keys": [
        {
            "alg": "RS256",
            "e": "AQAB",
            "kid": "63251aaf2369d6677f60faad64d2ca7031cf3ea5d0613c9be78feaf2ded5c477",
            "kty": "RSA",
            "n": "ztkZivJiKHTWB56SF9Qk6Ly4tt58mHwg6cDBkfYjte5j3k7ZgYlDLfsAubpAst4j3qujruU3GbWL4F4dDd7H93N8SxM1vckUggilEzsrKDxxX5wgHF0KPnubHqhgm88sfMef_4OT9S-RHKPr1PY_CTxKQOdUnsk0C28gn8dckxfUYRE4T6Cuk81NPoa4mke6ApB2kjtWBKctApB0kRedcufWfgb9W0pP9_EP0x8VELjTvTRFi7jXvRw5L_ImKODOhS-KrTSLTN0UOHMy8gaLc_4xH9wGvAoVe_bJI67ldeOft5XJ9g3v6YBhp7rcbnTZ2ksYMNLA4eX8qE0DWAj4ew==",
            "use": "sig"
        },
        {
            "kty": "EC",
            "alg": "ES256",
            "kid": "26df",
            "y": "xkUiCX5aPmZplJ5UVHk-DwWBywLEYcbweruF2qajDjU",
            "x": "DFYYgpiQbaBtxtlCm1ocw26xc_6tQGukFRVYCTK2x9M",
            "crv": "P-256",
            "use": "sig"
        }
    ]
}
```
So each backend deployed can have a public key listed in this file and you can have multiple backends as needed.

frontend/x509_scitokens_issuer/export-lib-symbols
```diff
@@ -0,0 +1,8 @@
+{
+global:
+  x509_scitokens_issuer*;
+  x509_macaroon_issuer*;
+
+local:
+  *;
+};
```
bbockelm: You don't need this file, this is just for the linker script of the C-based client.

frontend/deploy
bbockelm: Yes, I would suggest doing that at least in production.

It's very useful for the keys to be auto-generated in development (for pre-production - probably better to have the exact same procedure as production?)

So I'd leave the code here and just have Lina clobber the keys with the "real" ones on "deployment day".

frontend/deploy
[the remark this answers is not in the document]
bbockelm: Ok, I understand that part of the code now.

frontend/deploy
vkuznet: ok, in this case yes we'll generate keys elsewhere and copy to all frontend nodes such that they will have identical certs.

frontend/x509_scitokens_issuer/export-lib-symbols
vkuznet: done (removed)

PR #710, opened 2018-12-19
Added a DQM render plugin for CMS and LHC per lumisection status plot

andrius-k:
Added a render plugin for recently introduced CMS and LHC per lumisection status DQM plot. The plot was introduced in [this PR](https://github.com/cms-sw/cmssw/pull/25518).


andrius-k:
Hi @h4d4,

There will be an update to this PR so there is no reason to merge it right now.

amaltaro:
Should we close it?

andrius-k:
@amaltaro Yes, it's no longer necessary.

PR #711, opened 2018-12-19
Proof-of-concept for cmsweb + SSO

vkuznet:
This PR contains files required to enable CERN SSO in cmsweb. This work is mostly based on this blog [1] and contains proof-of-concept working setup for cmsweb and CERN SSO (Shibboleth) authentication. In order to work the following libs should be installed on frontend nodes:
```
sudo yum install -y shibboleth log4shib xmltooling-schemas opensaml-schemas
```
It provides `/usr/lib64/shibboleth/mod_shib_24.so` apache module used in frontend server configuration.

Also, before setting up frontend we should register once a new SSO application at
https://sso-management.web.cern.ch/

The request flow is the following:
- client access `/app` end-point
- client is redirected to CERN SSO if no credential cookies is found via `/login` end-point
- upon authentication CERN SSO `/login` is redirected back to `/app`
- the request is routed via ProxyPass to app URL

In my setup I run simple flask app on localhost:5000
```
# Flask app code
from flask import Flask, session, request

app = Flask(__name__)

@app.route('/')
def hello_world():
    kwds = request.args
    return 'Hello from authenticated app session={} kwds={}'.format(session, kwds)

if __name__ == '__main__':
    app.run(host='0.0.0.0')
```
This code was running on frontend node, and only represent a backend app. Actual cmsweb apps, such as DBS, ReqMgr2, etc. will be run on different backends and proposed configuration should be adopted accordingly (my guess we may need to use Proxy balancer in this case, see appropriate comment in `sso/sso_http.conf`).

At the moment I didn't find a way to use RewriteEngine rules we use currently since Shiboleth redirection points back to static web page rather then a service URI.

[1] https://alexpearce.me/2014/10/setting-up-flask-with-apache-and-shibboleth/


bbockelm:
Valentin, how would this integrate with the `cmsauth` authentication?  Is it possible to bypass the SSO integration if the user agent is already authenticated (e.g., via client certificate)?

vkuznet:
This is pure SSO authentication for web clients (I didn't yet change the port).
My understanding that we want to move all web clients to SSO auth schema (this
is what I'm trying with this PR) and all others (x509) clients will use
dedicated port (8443) which will still use `cmsauth` authentication schema
(which we need to re-write in python).

So, to answer your question, it would be possible for clients with x509 to
continue to use cmsauth schema, they will connect on port 8443, while web
clients will use default port (443) and use SSO schema.

I tried to use SSO+Flask+WSGI and it works to a certain degree, but there are
some obstacles which would be better to discuss in private (e.g. the redirected
uri changes and I can't find a way to keep it persistent after authentication).

PR #716, opened 2019-01-08
Add support to parse client certs from traefik headers

vkuznet:


vkuznet:
In order to use C libraries in Perl we need to rely either in Inline-C/Inline-CPP or perlxs. The former is much easier but tricky to add to apache, the latter I didn't try yet (but it seems much more complex approach which will require writing additional perl subroutines, compile code and deploy it).

In order to use simple-proxy-utils we need to add to the node the following set of RPMs:
**cmake, voms-devel, globus-gsi-credential-devel, globus-gsi-cert-utils-devel, globus-common-devel, globus-gsi-sysconfig-devel and globus-gsi-callback-devel**
Then, we need to adjust CMakeList.txt file and add the following flag: `-D_GLIBCXX_USE_CXX11_ABI=0`
After compiling the code we'll get libSimpleProxyUtils.so library which we may need to copy to a directory accessible by apache server, e.g. /data/srv/state/frontend/libs/ and also need to create over there /data/srv/state/frontend/libs/Inline directory for Inline-C to store compiled code.

After that we may add Inline C call into cmsauth (see this PR) to invoke call to desired C library API call.

References:
- https://metacpan.org/pod/distribution/Inline-C/lib/Inline/C/Cookbook.pod
- https://github.com/ingydotnet/inline-c-pm
- https://metacpan.org/pod/distribution/Inline-C/lib/Inline/C.pod
- https://metacpan.org/pod/release/DAVIDO/Inline-CPP-0.39/lib/Inline/CPP.pod
- http://perldoc.perl.org/perlxs.html
- https://github.com/bbockelm/simple-proxy-utils/blob/master/src/SimpleProxyUtils.cc

PR #719, opened 2019-01-21
Hg1902 cric sec mod

h4d4:
This PR:
- Checks that fetched json has valid fields in the right order = DN, ID, LOGIN, NAME, ROLES. 
- Always set in lower case ROLES values, does not matter if those already are  in.

### DIFF
--- a/frontend/mkauthmap
+++ b/frontend/mkauthmap
@@ -111,6 +111,30 @@ def check_keys(jsdata):
 
     return True
 
+def roles_to_lower(jsdata):
+    data = json.loads(jsdata)
+
+    for entry in data:
+        roles2 = {}
+        for role in entry['ROLES']:
+            cur_key = role
+            cur_val = entry['ROLES'][cur_key]
+
+            cur_key = role.lower()
+
+            for i in range(0, len(cur_val)):
+                cur_val[i] = cur_val[i].lower()
+
+            roles2[cur_key] = cur_val
+        
+        del entry['ROLES']
+        entry['ROLES'] = roles2
+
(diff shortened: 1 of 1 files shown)


amaltaro:
@h4d4 Lina, can you please give me one example of how is the data now and how you want it to be?
I think we can make this code simpler and more readable, I just want to make sure I understand what exactly you want.

h4d4:
@amaltaro Alan, the mains  goal with the PR, are:

- https://github.com/dmwm/deployment/blob/d517f41a2b59f1a82445ac048b0b8e93b8ef8be4/frontend/mkauthmap#L183
ensure that old json content only will replaced with the new json content if, the new json contend keeps the correct structure, that is:
`fields = DN, ID, LOGIN, NAME, ROLES`
as for example:
 {"DN": "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=amaltaro/CN=718748/CN=Alan Malta Rodrigues","LOGIN": "amaltaro", "NAME": "Alan Malta Rodrigues", "ROLES": {"admin": ["group:phedex", "group:reqmgr"], "dbsexpert": ["group:dbs"], "operator": ["group:dbs"], "production-operator": ["group:dataops"], "user": ["group:users"]}},

If new json content has less, or more fiels that the defined(`fields` ), or in different order, oldjson content not will be updated.

- always set ROLES in lower case, does not matter if this if already on lower case

https://github.com/dmwm/deployment/blob/d517f41a2b59f1a82445ac048b0b8e93b8ef8be4/frontend/mkauthmap#L184

This is done, after PHEDEX roles extension to have this in isolated way, thinking in that probably this will be remove in some point.

PR #720, opened 2019-01-22
Validate keys in the mapping; make roles/groups lower case

amaltaro:
Possibly superseeding #719

Real changes are in the 1st commit. 2nd commit only contains aesthetic changes to make the code a bit more readable.

As discussed with Lina, this PR contains the following:
* convert group and roles information to lower case (even though CRIC is supposed to return it already in lower case)
* make sure the json content only contains supported attributed, so validate all the keys before persisting it.

@h4d4 besides checking whether this works or not, can you also check whether the order of attributes matter or not?

### DIFF
--- a/frontend/mkauthmap
+++ b/frontend/mkauthmap
@@ -98,6 +98,29 @@ def buildSiteMap(sites):
     return sitemap
 
 
+def rolesToLower(dataDict):
+    """
+    Receives a python dictionary with all the user entries and convert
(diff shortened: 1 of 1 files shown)


h4d4:
@amaltaro Alan, thanks for the contributions! 
Regarding attributes order I'll additionally define a test case in dev environment, to ask for example crab team if they face problem with the authentication.
For know, I don't see showstoppers to merge the PR.

Best Regards, Lina.

PR #725, opened 2019-02-21
Frontend reboot

h4d4:
This PR modified frontend/manage in order to restart frontend service as _frontend user, once the node is rebooted an below cron entry is trigered:
@reboot sudo bashs -l -c "/data/srv/current/config/frontend/manage restart 'I did read documentation'"


amaltaro:
Just wondering if it isn't cleaner to just not restart the frontend at reboot time(?) 
This is how we had CMSWEB for many years and I was never aware of any problems caused by the lack of that cron entry.

PR #733, opened 2019-03-01
Further improvement to frontend manage script

vkuznet:
I further simplified frontend manage script, changes include the following:

- add cleanup function which takes care of killing orphan httpd processes and cleaning semaphor files
- changes to stop/start/restart actions
- stop httpd daemon processes only if they are present in a system

### DIFF
--- a/frontend/manage
+++ b/frontend/manage
@@ -31,21 +31,43 @@ if [ -f /usr/sbin/apachectl ]; then
     sudo /usr/sbin/apachectl stop
 fi
 
-case ${1:-status} in
-  status | graceful | stop | configtest )
-    $STATEDIR/etc/httpd $1
-    ;;
-
-  start | restart )
-    echo "stopping httpd service ..."
-    $STATEDIR/etc/httpd stop
+cleanup()
+{
+    pids=`ps auxww | grep http | grep frontend | awk '{ORS=" "; print $2}'`
+    if [ -n "$pids" ]; then
+        sudo kill -9 $pids
+    fi
     ipids=`ipcs -s | awk '/apache|frontend/ {print $2}'`
     if [ -n "$ipids" ]; then
         echo "for i in $ipids; do (sudo ipcrm -s $i); done"
-        for i in `ipcs -s | awk '/apache/ {print $2}'`; do (sudo ipcrm -s $i); done
+        for i in `ipcs -s | awk '/apache|frontend/ {print $2}'`; do (sudo ipcrm -s $i); done
     fi
-    echo "restarting httpd service ..."
-    $STATEDIR/etc/httpd restart
+}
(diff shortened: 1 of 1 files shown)


vkuznet:
This PR contains changes to cmsweb frontend manage script to fix frontend start-up issues with crontab upon node reboot. But additional changes are required to CC7 node:

- we need to disable system httpd service, here is required set of commands
```
sudo systemctl disable httpd.service
systemctl status httpd.service
```
- we should ensure that user account allows password-less execution of sudo commands. This can be achieved by the following
```
# replace <USER> with actual user name where cronjobs will be installed
sudo cp /etc/sudoers.d/000-sudo-users /etc/sudoers.d/000-sudo-<USER>
# then we should modify 000-sudo-<USER> file and add this line:
<USER>      ALL=(ALL)       NOPASSWD:ALL
# please be sure it should not contradict to other rules
```

Further references can be found here:

1. https://www.cyberciti.biz/faq/check-running-services-in-rhel-redhat-fedora-centoslinux/
2. https://linuxconfig.org/how-to-stop-start-and-disable-enable-firewall-on-redhat-7-linux-system

h4d4:
@vkuznet Valentin,

I have not clear below section:
> - we should ensure that user account allows password-less execution of sudo commands. This can be achieved by the following
> ```
> # replace <USER> with actual user name where cronjobs will be installed
> sudo cp /etc/sudoers.d/000-sudo-users /etc/sudoers.d/000-sudo-<USER>
> # then we should modify 000-sudo-<USER> file and add this line:
> <USER>      ALL=(ALL)       NOPASSWD:ALL
> # please be sure it should not contradict to other rules
> ```
which user is referencing by: `user account `: `cmsweb` or `frontend `?
I've have checked nodes and there is not this file: `/etc/sudoers.d/000-sudo-users`
below is what we have:
```
[cmsweb@vocms0734 ~]$ sudo ls -l /etc/sudoers.d/
total 12
-r--r-----. 1 root root 316 May  9  2018 cmsweb
-r--r-----. 1 root root  33 May  9  2018 cmswebconf
-r--r-----. 1 root root 173 Mar 11 12:48 superusers
[cmsweb@vocms0734 ~]$ 
```
And /etc/sudoers.d/cmsweb has the structure pointed above: `<USER>      ALL=(ALL)       NOPASSWD:ALL` including all user services.

Then we also need this additional configuration?

Best Regards, Lina

vkuznet:
Lina, if /etc/sudoers.d/cmweb has all service users then it is fine. I'm referring users who will need sudo privileges, I think in this case both cmsweb and frontend. In my case when I tested on my VM I granted sudo to myself (user valya). If you're testing it under cmsweb account then this account should be used.

PR #735, opened 2019-03-06
Fix das2go exporter address parameter; add sysboot action

vkuznet:


exporters/deploy
```diff
@@ -20,4 +20,8 @@ deploy_exporters_post()
 {
   # keep this line to avoid error on empty function
   local cmd="test"
+  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740) disable ;; * ) enable ;; esac
```
h4d4: @vkuznet Valentin, cross-checking with nodes defined in manage script, this list of nodes in which the service wont be enabled ('disabled'), is including some nodes that should not be. 

So, below current cmsweb list of nodes:
production frontends: vocms0{158,760,162,164}
production backends: vocms0{136,161,163,165,738,739,740,741,742}
testbed frontends:  vocms0{734,135}
testbed backends:  vocms0{731,132}

Therefore taking in account above  and  manage script I'd defined the list such:

`  case $host in vocms073[89] | vocms0740) disable ;; * ) enable ;; esac`

PR #736, opened 2019-03-07
Add new enable/disable rule; make sysboot a separate action

vkuznet:


exporters/deploy
```diff
@@ -1,14 +1,13 @@
 # vim: set ft=sh sw=2 ts=8 et :
 deploy_exporters_deps()
```
h4d4: @vkuznet Valentin, I think that here we have an error of concept regarding if the service is a fronted or a backed flavor. I'm making reference to below functions:
```
deploy_exporters_prep
deploy_exporters_deps
```
What I see in that in principle, this service was created with specifications for a backed service flavour,
 for example, checking deployment scripts for all backends services, all they includes `$stage backend` in deploy_exporters_deps method.

On another hand, checking deployment script for the only one 'fronted' service that we have until now, [deploy_exporters_deps](https://github.com/dmwm/deployment/blob/master/frontend/deploy#L3) definition is so different.

I'm going to merge and test it,  and made the the necessaries changes by hand if those are necessaries.
Until now I've been successfully deployment the service as a backed, the problems came up when I thy to deploy this same service in fronted nodes.
I'll let you know what I found.

PR #743, opened 2019-03-26
Add oomscore to list of monitoring metrics

vkuznet:


admin/ServerMonitor
```diff
@@ -193,6 +193,24 @@ class SystemMonitor(Monitor):
           addfield(data, field, safefield('io', proc.io_counters))
         elif field == 'files':
           addfield(data, field, safefield('files', proc.open_files))
+        elif field == 'oomscore':
+          try:
+            oom_score = int(open('/proc/%s/oom_score' % proc.pid).read().replace('\n', ''))
```
amaltaro: Valentin, use a context manager please. Otherwise you cannot be sure whether the file object was properly closed or not. In other words, instead of
`oom_score = int(open('/proc/%s/oom_score' % proc.pid).read().replace('\n', ''))`
use something like (please check the syntax)
```
with open('/proc/%s/oom_score' % proc.pid) as fo:
    oom_score = int(fd.read().replace('\n', ''))
```

Same for the lines below.

vkuznet:
Alan, you're right, I made appropriate adjustment.

PR #746, opened 2019-03-28
Fix value of added field

vkuznet:


yuyiguo:
Valentin,

This PR fixed the pids are all 1, right?
Yuyi

vkuznet:
Yuyi, it should be underlying process PID, see
https://github.com/dmwm/deployment/blob/a95061eb37f58f40995491c90487e6ad89b09c23/admin/ServerMonitor#L160
which loops over all processes in proc file system and then here I just read and store PID:
https://github.com/dmwm/deployment/blob/a95061eb37f58f40995491c90487e6ad89b09c23/admin/ServerMonitor#L197

yuyiguo:
Thanks Valenint.
@h4d4 Could you please patch the cmsweb-testbed with this PR?  Thanks, Yuyi

h4d4:
@yuyiguo Yuyi, it was already done, such is shown [here](https://gitlab.cern.ch/cms-http-group/doc/issues/149#note_2487278) .
The changes were set by hand because the idea test first the small changes before run a whole testbed deployment.

Also, see please current one value for line 160:

```
[cmsweb@vocms0731 admin]$ sed -n '160p' /data/srv/current/config/admin/ServerMonitor
    for proc in psutil.process_iter():
[cmsweb@vocms0731 admin]$
```

Best Regards, Lina.

yuyiguo:
Thanks Lina for confirming the patch was applied.
@vkuznet 
Valentin,
Below is the out of server-monitor-20190329.log after Lina applied the patch. The PID is first one and it was still 1. It also has the message of "failed to get files ... psutil.AccessDenied " just before each of the  "_dbs SYSTEM PROCESSES"  line. 
Any idea about this?

```
Mar 29 19:11:19 _dbs SYSTEM IGNORE failed to get files field information: psutil.AccessDenied (pid=6247)
Mar 29 19:11:19 _dbs SYSTEM PROCESSES: 1, CPU: 0.1%, MEM: 0.8%, FILES: 0, OOMSCORE: 4, OOMADJ: 0, OOMSCOREADJ: 0, SWP: 0.0%, THREADS: 18, USER: 89.66, SYSTEM: 18.06, RSS: 124579840, VMS: 551985152, SWAP: 1986560
```

vkuznet:
Yuyi,
the SYSTEM PROCESSES: 1 means that it is only 1 process, but it is not a pid.
Similar message appears in other monitoring service records, e.g. sitedb,
mongodb, etc. The '1' has nothing to do with PID.

Turns out, see [1], the ServerMonitor has pid as reserved keyword and it is
skipped from the output in logs. The pid refers to type of monitoring rather
then pid of the processes. To minimize the changes and do not break existing
functionality for other services which may use pid type monitoring,
I decided to rename pid in process activity to procid and created necessary PR: [2].
It means that you need to change your monitoring-*.ini files to use procid,
instead of pid, i.e. instead of
PROCESS_ACTIVITY="pid,cpu,mem,files,oomscore,oomadj,oomscoreadj,swp,threads,user,system,rss,vms,swap"
use
PROCESS_ACTIVITY="procid,cpu,mem,files,oomscore,oomadj,oomscoreadj,swp,threads,user,system,rss,vms,swap"
with [1].

But second issue 
> Mar 29 19:11:19 _dbs SYSTEM IGNORE failed to get files field information: psutil.AccessDenied (pid=6247)
related to the fact that Server Monitor runs under cmsweb account and access to
file descriptor info requires root privilege. That's why FILES: X did't not
appear in process activity output string. To fix the issue the ServerMonitor
should run with sudo.

Lina, could you please manually apply [2] to fix skipped process id, then we
need adjustements to DBS monitoring-*.ini files I mentioned above.
And, to fix FILES message we need to run ServerMonitor with sudo. I made
PR for that too [3], but I don't have time to test it. So if you can do it
manually and verify that everything works (check server monitor logs and
see if nothing breaks for dbs and other processes), then we can include it in
release.

Best,
Valentin.

[1] https://github.com/dmwm/deployment/blob/226169f464132a00ed40205e167ca5139359c0d8/admin/ServerMonitor#L225
[2] https://github.com/dmwm/deployment/pull/747
[3] https://github.com/dmwm/deployment/pull/748

What is your minitoring configuration? You should have it like this:

yuyiguo:
Valentin,

I think it may be a typo in the patch https://github.com/dmwm/deployment/pull/747/files#diff-a18a219bff653901021cdd9218825c99R196
addfield(data, field, safefield('procpid', proc.pid))

procpid should be procid?
Yuyi

vkuznet:
Try again, use procpid.

vkuznet:
closing this PR since I provided all changes in #747

PR #748, opened 2019-03-29
Adjust crontab entries to use sudo for ServerMonitor

vkuznet:


h4d4:
@yuyiguo @vkuznet Yuyi, Valentin

Just to confirm the merging. I've just saw that the the same pointed out in the email.

Best Regards, Lina.

PR #750, opened 2019-04-02
Use open files call only when metrics is specified 

vkuznet:
The files metrics requires sudo access and if used for all monitors it causes system failure.


muhammadimranfarooqi:
Hi @vkuznet 

This is old PR. Should I close this if it not needed.

vkuznet:
I still suggest to merge it as it fix issue in ServerMonitor which still in use in VMs deployment.

PR #761, opened 2019-04-24
Add other DBS instances to monitor

vkuznet:


yuyiguo:
@vkuznet @h4d4 
Here is the DBS instances' port assignment
8251: Migration
8252: global/DBSReader"
8253: global/DBSWriter
8254: phys03/DBSReader
8255: phys03/DBSWriter
8250: def

8250 is the default that has no traffic on it so no needs to be monitored.  It will be nice to have  the 1000+instance port schema of the exporters match the under line services.

PR #763, opened 2019-04-30
Setup JAVA_HOME for migrate2hdfs if it is missing

vkuznet:


muhammadimranfarooqi:
Hi @vkuznet 

This is old PR. Should I close this if it not needed.

vkuznet:
I still suggest to merge it as it fixes the issue with the script and we may still use this script in a future.

PR #765, opened 2019-05-01
New gp 766

h4d4:
This PR updates deployments scripts to include new general propose backend(gp) that will be called vocms0766

- Regarding commit '[648f895](https://github.com/dmwm/deployment/commit/648f895c685c469cdeec47dd25cd37e74aafb135) '  below is the list of `services*/deployment` where general propose backends are `'disable'`, therefore on those scripts the new gp need to be included:

```
#gps ARE LISTED TO BE 'disable' =>  below services are NOT DEPLOYED in gps nodes
confdb
couchdb
crabcache
das
dcafpilot (no deployed anymore in the cluster)
dqmgui
filemover
mongodb
overview (no deployed anymore in the cluster)
phedexreplicamonitoring
reqmgr2ms (until now just in testbed nodes)
tfaas
wmarchive (##updating their deployment script even when this is not deployed into cluster)
```
In that commit I've also included `exporters/manage`, the script made a explicity list of nodes where exporters runs, and there gps are listed.

-  I've not modified below  `services*/deployment` scripts due to gps are not listed to be  'disable', it means that those services are deployed into gps, then is not necessary update it:
```
###SERVICES WHERE GPS ARE no LISTED TO BE 'disable' => below services are DEPLOYED in GPS nodes ##deployment updates, is not necessary
crabserver
dbs
dbsmigration
dbsweb (no deployed anymore in the cluster)
dmwmmon
exporters
gitweb (no deployed anymore in the cluster)
phedex
popdbweb
reqmgr2 
reqmon
sitedb
t0_reqmon = CMS T0 WMStats (T0ReqMon)
t0wmadatasvc
victorweb
```
`NOTE: ` 
In the end, what decides if x deployed service process queries or no, are `fronted-rules` [(backends-prod.txt)](https://github.com/dmwm/deployment/blob/master/frontend/backends-prod.txt). So, there are cases where one service is deployed in several cmsweb's nodes but service's requests are redirected to certain of those nodes  by `fronted-rules`  
Looks like that configuration of deploy services in nodes where finally request will no be handled is part of the way in which the cluster was designed.

- I've not included the commit for `frontends-backends rules` because, it need to be done once the node is complete functionally.

### DIFF
--- a/admin/deploy
+++ b/admin/deploy
@@ -35,7 +35,7 @@ deploy_admin_post()
 
   local pxlabel pxskip certs
   case $host in
-    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms0760 | vocms030[67] | vocms0318 | vocms0766)
+    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms076[06] | vocms030[67] | vocms0318 )
       pxlabel=cmsweb_backends ;;
     * )
       pxlabel=devvm_$(echo $host | tr - _) ;;
--- a/admin/deploy
+++ b/admin/deploy
@@ -35,7 +35,7 @@ deploy_admin_post()
 
   local pxlabel pxskip certs
   case $host in
-    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms0760 | vocms030[67] | vocms0318 )
+    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms0760 | vocms030[67] | vocms0318 | vocms0766)
       pxlabel=cmsweb_backends ;;
     * )
       pxlabel=devvm_$(echo $host | tr - _) ;;
--- a/das/deploy
+++ b/das/deploy
@@ -33,7 +33,7 @@ deploy_das_sw()
 
 deploy_das_post()
 {
-  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740 | vocms0766) disable ;; * ) enable ;; esac
+  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740 | vocms0766 ) disable ;; * ) enable ;; esac
   (mkcrontab
    sysboot
    for action in das_cleanup:'0 0-23/2 * * *'; do
--- a/das/deploy
+++ b/das/deploy
@@ -33,7 +33,7 @@ deploy_das_sw()
 
 deploy_das_post()
 {
-  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740) disable ;; * ) enable ;; esac
+  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740 | vocms0766) disable ;; * ) enable ;; esac
   (mkcrontab
    sysboot
    for action in das_cleanup:'0 0-23/2 * * *'; do
--- a/exporters/manage
+++ b/exporters/manage
@@ -82,7 +82,7 @@ start()
     # production nodes
     #
     # DBS on production nodes
-    vocms0136 | vocms0161 | vocms0163 | vocms0165 )
+    vocms0136 | vocms0161 | vocms0163 | vocms0165 | vocms0766 )
         # start process_exporter for DBS server
         wait4proc "DBSGlobalReader" # check for pattern that DBS is running
         process_monitor.sh ".*DBSGlobalReader" dbs_global_exporter ":18252" 15 \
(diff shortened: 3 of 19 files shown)


vkuznet:
Lina, changes looks fine to me. I understand the logic. But as I wrote in private email we'll need to adjust backend-prod.txt to explicitly add DBS redirect rules to certain nodes (current 4 nodes plus new backend 0766) and have default to current 4 nodes for everybody else.

admin/deploy
```diff
@@ -35,7 +35,7 @@ deploy_admin_post()
 
   local pxlabel pxskip certs
   case $host in
-    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms0760 | vocms030[67] | vocms0318 )
+    vocms0117 | vocms0127 | vocms013[245689] | vocms0734| vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms0760 | vocms030[67] | vocms0318 | vocms0766)
```
amaltaro: We can match 0760 and 0766 with a regex, so we could have something like (and space is needed before closing parenthesis AFAIK):
```
vocms0117 | vocms0127 | vocms013[245689] | vocms0734 | vocms073[189] | vocms0143 | vocms074[012] | vocms016[01235] | vocms076[06] | vocms030[67] | vocms0318 )
```

das/deploy
```diff
@@ -33,7 +33,7 @@ deploy_das_sw()
 
 deploy_das_post()
 {
-  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740) disable ;; * ) enable ;; esac
+  case $host in vocms013[689] | vocms073[89] | vocms016[135] | vocms0740 | vocms0766) disable ;; * ) enable ;; esac
```
amaltaro: blank space needed here

exporters/manage
```diff
@@ -120,7 +120,7 @@ start()
       ;;
 
     # Apache exporter on production nodes
-    vocms0158 | vocms0760 | vocms0162 | vocms0164 )
+    vocms0158 | vocms0760 | vocms0162 | vocms0164  | vocms0766 )
```
amaltaro: You don't need it here (only frontend nodes)

exporters/manage
```diff
@@ -225,7 +225,7 @@ status()
         process_status "couchdb_exporter"
       ;;
 
-    vocms0158 | vocms0760 | vocms0162 | vocms0164 )
+    vocms0158 | vocms0760 | vocms0162 | vocms0164  | vocms0766 )
```
amaltaro: Same here, please remove.

h4d4:
@bbockelm @yuyiguo @vkuznet Valentin, thanks very much for check it.
Regarding [backends-prod](https://github.com/dmwm/deployment/blob/master/frontend/backends-prod.txt), I need to have it clear from all you experts, because what I understood for @bbockelm commets[0] is that this new gp will also serving the others services as well as the current gps on production, but the update here will be double RAM.
Therefore, to reach above the change will be add `vocms0766` to this line [line](https://github.com/dmwm/deployment/blob/master/frontend/backends-prod.txt#L33).

Brian please let me know your comments.

Best Regards Lina.
[0] Brian's comments: 
"I'm suggesting upgrading one (so we have N+1 capacity), then sequentially touching the others.  This way, we always maintain N DBS backends and never N-1 backends.  We should upgrade the memory on all of them while we're at it."

amaltaro:
I'm not sure I got what Valentin wanted to say, but I think he said the same as Brian and the same as everyone understood, for the moment.
We add a new node to CMSWEB - which can serve any GP requests, thus running any GP services - then the other nodes start going down one by one, which means we will always have 4 GP backends up & running.

Having a specific backend for a specific client is out of the scope for this very near future, AFAICT.

vkuznet:
Lina,
yes, the change would be to add `vocms0766` to last line.

admin/deploy
h4d4: Yes, it true.

das/deploy
h4d4: the same, I'll add the space.

exporters/manage
h4d4: yes, thanks! it is a mistaken.

exporters/manage
h4d4: yes, the same error.

h4d4:
@amaltaro Alan,

thanks for the review, I'll apply the changes and let you know. 

Best Regards, Lina.

h4d4:
@amaltaro Alan, 
I've just applied above pointed  corrections.

amaltaro:
It looks good now, Lina. However, if it was a PR against WMCore, I'd never merge it because it has 9 commits! Given it's under the deployment repo, you do as it pleases you :), or you spend < 5 and squash all those commits in a single one, anyone looking at the history of those files will be happy to see a single commit instead of commits back and forth.

h4d4:
@amaltaro Alan, 
thanks for the comments I'll keep in mind it to improve :) 
Yes since it's deployment repo I'm responsible to merge it.
I'll review git doc for squashing 

Best Regards, Lina.

amaltaro:
@h4d4 just in case you missed this PR, I believe you wanted to push it to testbed in HG1906, no?

h4d4:
@amaltaro Alan, 

Thanks for check it. nevertheless, this PR was overseed by PR #766.
I'll close this one.

h4d4:
Closing PR, since it was overseed by PR #766.

PR #768, opened 2019-05-13
double threads for Globalreader

yuyiguo:


amaltaro:
Did you really wanted to merge it? I thought the plan was to evaluate how the service would behave with those 30 threads, before making it standard.

h4d4:
@yuyiguo Yuyi, If you agree we can revert this PR, until this change be an standard.
Otherwise, all deployments that we run including it, will set this value by default.

Please let me know.

Best Regards, Lina.

PR #769, opened 2019-05-14
update db connection pool to 30 as the cherrypy threads.

yuyiguo:


h4d4:
@yuyiguo Merging PR to ensure value in next deployments.

PR #771, opened 2019-05-16
Set CherryPy accepted_queue_size and cpstats monitoring for a couple of services

amaltaro:
Just a placeholder to show how this parameter has to be set in some of our applications.
FYI @belforte @yuyiguo

### DIFF
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -28,6 +28,9 @@ config = Configuration()
 main = config.section_("main")
 srv = main.section_("server")
 srv.thread_pool = 30
+# The maximum number of requests which will be queued up before
+# the server refuses to accept it (default -1, meaning no limit).
+srv.accepted_queue_size = -1
 main.application = "reqmgr2"
 main.port = 8246 # main application port it listens on
 main.index = "ui"
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -31,6 +31,7 @@ srv.thread_pool = 30
 # The maximum number of requests which will be queued up before
 # the server refuses to accept it (default -1, meaning no limit).
 srv.accepted_queue_size = -1
+srv.accepted_queue_timeout = 0
 main.application = "reqmgr2"
 main.port = 8246 # main application port it listens on
 main.index = "ui"
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -40,8 +40,10 @@ main.authz_defaults = {"role": None, "group": None, "site": None}
 #set default logging (prevent duplicate)
 main.log_screen = True
 
-sec = main.section_("tools").section_("cms_auth")
-sec.key_file = "%s/auth/wmcore-auth/header-auth-key" % ROOTDIR
+tools = main.section_("tools")
+# provide CherryPy monitoring under: <hostname>/reqmgr2/data/stats
+tools.section_("cpstats").on = False
+tools.section_("cms_auth").key_file = "%s/auth/wmcore-auth/header-auth-key" % ROOTDIR
 
 # this is where the application will be mounted, where the REST API
 # is reachable and this features in CMS web frontend rewrite rules
--- a/crabserver/config.py
+++ b/crabserver/config.py
@@ -9,6 +9,9 @@ conf = Configuration()
 main = conf.section_('main')
 srv = main.section_('server')
 srv.thread_pool = 25
+# The maximum number of requests which will be queued up before
+# the server refuses to accept it (default -1, meaning no limit).
+srv.accepted_queue_size = -1
 main.application = 'crabserver'
 main.port = 8270
 main.index = 'data'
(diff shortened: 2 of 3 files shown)


yuyiguo:
@amaltaro 
Alan, What happens when accepted_queue_size is reached?

amaltaro:
Cherrypy server will start refusing requests when that's reached, see from the documentation:
https://github.com/cherrypy/cherrypy/blob/ce51624d23208bbeffddf37de84a3238f64ede7f/cherrypy/_cpserver.py#L62

not sure about which error is returned to the client though. If you're willing to test it, you can set the `thread_pool=1`, set `accepted_queue_size=0`, and run 2 or 3 requests in parallel (or slow requests almost in parallel).

reqmgr2/config.py
```diff
@@ -28,6 +28,9 @@
 main = config.section_("main")
 srv = main.section_("server")
 srv.thread_pool = 30
+# The maximum number of requests which will be queued up before
```
belforte: can you put this enlightening comments also in crabserver/config.py . thanks

bbockelm:
FWIW - if this is hit, CherryPy will abruptly close the TCP socket even before processing the HTTP headers.  This will cause the Apache frontend to respond with a proxy error (I think it does a 502 these days).

It's at least theoretically possible to send back a nicely-formatted message from within the WMCore framework, but I have a hard time convincing myself it's worth the effort (if we spend time here, we should spend time getting rid of CherryPy...).

That said, we _should_ have the ability to pull statistics into Prometheus on how often this event occurs.

amaltaro:
I made further updates to this PR such that we can also enable/disable the CherryPy statistics (`tools.cpstats.on`), as provided by @vkuznet here:
https://github.com/dmwm/WMCore/pull/9205

It's all disabled now, so even if we merge it, there should be no impact.

About the `accepted_queue_size`, I did some quick tests and indeed the client receives a `502 Proxy Error`.  One thing that caught my attention though is that the proxy error doesn't come right away; it seems to rely on another CP parameter called `accepted_queue_timeout`. I think we should set that one to either 0 or 1, such that it doesn't hold frontend apache requests, right?

amaltaro:
Configuration updated to drop the request right away, in case the cherrypy queue size has reached the limit (`accepted_queue_timeout=0`).

yuyiguo:
DBS does not set  accepted_queue_size currently. I guess by default it is  -1. So the 502 errors of DBS would not come from the queue?  Could the unlimited queue cause trouble in FE?

If we want to pull statistics into Prometheus for cherrypy, we need to turn cpstats to ture.  Are we going to have the statistics?

amaltaro:
This pull request provides the exact configurations that need to be changed, for now I'm setting all of those to their default values. 

Yes, having an unlimited queue size means that requests are queued on the backend, meaning it also allocates an apache thread and reduces the overall amount of requests that can go in in the whole cmsweb cluster.

> If we want to pull statistics into Prometheus for cherrypy, we need to turn cpstats to ture. Are we going to have the statistics?

I think Valentin needs to wait for one of the services to start providing it, then he can work on a dashboard for that. The monitoring overhead is minimal, but not 0, so you might check whether it has any other side-effects when enabled.

vkuznet:
Yuyi, to get stats of Cherrypy in Prometheus two things should happen:
- someone need to turn on the configuration for a service (e.g. DBS)
- I need to write new exporter for this stats
Once both items will be completed the Prometheus will get stats and we can build
dashboard.

Therefore, if you want to have cherrypy stats for DBS you need to turn the
configuration flag and deploy it in new cmsweb release.

yuyiguo:
Valentin,
I am thinking may be we can  turn on cpstats monitoring for  DBSPhys03Writer in June release as a release. This instance has lower access rate than others.
Yuyi

vkuznet:
That's fine by me. But as I said before I need to write new exporter for that 
in order to get metrics in Prometheus. I'm working on it now.

amaltaro:
@yuyiguo @belforte if you agree, I'll ask Lina to merge this PR tomorrow.
I defined all default settings - so all these changes should be completely transparent - but this one:
` srv.accepted_queue_timeout = 0`

default time is 10 secs. However, since we define unlimited queue_size, this change should have no effect at all.

With this, I'll get those two CherryPy PRs merged in WMCore as well.

yuyiguo:
Alan,
I am working on what DBS’s queue size should. A lot of thing are going at the same time.
Yuyi

belforte:
huh.. I did not understood that this was involving me as well, so I payed no attention.
do I really need to understand ? I am all day in other things today

amaltaro:
Stefano, Yuyi, this was just a placeholder such that you know where and how exactly those parameters have to be set (I believe they will eventually land on your to-do list...).

I made another PR specific only to ReqMgr2:
https://github.com/dmwm/deployment/pull/774

so I'm closing this one and you two can use it for reference in the future.

yuyiguo:
Alan,
It is fine to close it for now.
Yuyi

PR #783, opened 2019-06-04
Enable strict mode in MariaDB; change file format to Barracuda; enable large prefix; other tweaks

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/9242

Apparently we were not using the strict mode in MariaDB since the migration from MySQL (different syntax in the config file).
Option `tx_isolation` was completely removed, apparently we only depend on `transaction-isolation`.

A few other options were removed/updated according to the MariaDB logs:
```
2019-06-04 18:07:24 140249885366400 [Warning] No argument was provided to --log-bin and neither --log-basename or --log-bin-index where used;  This may cause repliction to break when this server acts as a master and has its hostname changed! Please use '--log-basename=vocms0260.cern.ch' or '--log-bin=mysqld-bin' to avoid this problem.
2019-06-04 18:07:24 140249885366400 [Note] Using unique option prefix 'key_buffer' is error-prone and can break in the future. Please use the full name 'key_buffer_size' instead.
2019-06-04 18:07:24 140249885366400 [Warning] 'THREAD_CONCURRENCY' is deprecated and will be removed in a future release.
2019-06-04 18:07:25 7f8e7898d880 InnoDB: Warning: Using innodb_additional_mem_pool_size is DEPRECATED. This option may be removed in future releases, together with the option innodb_use_sys_malloc and with the InnoDB's internal memory allocator.
2019-06-04 18:07:25 7f8e7898d880 InnoDB: Warning: Setting innodb_checksums to OFF is DEPRECATED. This option may be removed in future releases. You should set innodb_checksum_algorithm=NONE instead.
2019-06-04 18:07:25 7f8e7898d880 InnoDB: Warning: Using innodb_locks_unsafe_for_binlog is DEPRECATED. This option may be removed in future releases. Please use READ COMMITTED transaction isolation level instead, see http://dev.mysql.com/doc/refman/5.6/en/set-transaction.html.
```

### DIFF
--- a/wmagent/my.cnf
+++ b/wmagent/my.cnf
@@ -1,18 +1,18 @@
 [mysqld]
-max_allowed_packet=128M
-transaction-isolation = READ-COMMITTED
+# only STRICT_TRANS_TABLES mode is not default in MariaDB 10.1.x
+sql_mode="NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,STRICT_TRANS_TABLES"
+transaction-isolation=READ-COMMITTED
 
+max_heap_table_size=2048M
+max_allowed_packet=128M
 max_connections = 200
 connect_timeout = 60
 
-innodb_log_file_size=512M
-innodb_log_buffer_size=8M
-innodb_buffer_pool_size=2G
-innodb_additional_mem_pool_size=50M
-#set-variable = transaction-isolation = READ-COMMITTED
 binlog_format=row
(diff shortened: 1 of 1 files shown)


amaltaro:
@h4d4 Lina, I tested these changes and they look Ok. Can you please merge it? Thanks

h4d4:
@amaltaro Alan, 
I've just merged it.

Best Regards, Lina

PR #792, opened 2019-07-11
Start DBS instance if its configuration file is present

vkuznet:
@yuyiguo , @h4d4 please review and merge this patch, since this is useful for separating DBS instances in k8s setup, i.e. we can start individual DBS instance per k8s pod/node by using different configurations. For example, we can create configuration for individual dbs instances and then apply the same manage script to start individual DBS instances.

### DIFF
--- a/dbs/manage
+++ b/dbs/manage
@@ -100,12 +100,24 @@ start_dbs()
 
 start()
 {
-  start_dbs $CFGFILE
-  start_dbs $CFGFILE_GR
-  start_dbs $CFGFILE_GW
-  start_dbs $CFGFILE_P3R
-  start_dbs $CFGFILE_P3W
-  start_dbs $CFGFILE_MG
+    if [ -f $CFGFILE ]; then
+        start_dbs $CFGFILE
+    fi
+    if [ -f $CFGFILE_GR ]; then
+        start_dbs $CFGFILE_GR
+    fi
+    if [ -f $CFGFILE_GW ]; then
+        start_dbs $CFGFILE_GW
+    fi
+    if [ -f $CFGFILE_P3R ]; then
+        start_dbs $CFGFILE_P3R
+    fi
+    if [ -f $CFGFILE_P3W ]; then
+        start_dbs $CFGFILE_P3W
+    fi
+    if [ -f $CFGFILE_MG ]; then
+        start_dbs $CFGFILE_MG
+    fi
 }
 
 # Stop the service.


yuyiguo:
@vkuznet 
Hi Valentin,

How the configuration file will be deployed in k8s?  Here is what in the dbs/manage:

CFGFILE_GR=DBSGlobalReader.py
CFGFILE_GW=DBSGlobalWriter.py
CFGFILE_P3R=DBSPhys03Reader.py
CFGFILE_P3W=DBSPhys03Writer.py
CFGFILE_MG=DBSMigrate.py

Above files are all in the same directory as dbs/manage.  
 In k8s development, I would think there would be only one cfg file per container. So we need to check which one is there before we assign CFGFILE_* , then we will only assign one of them to the corresponding file 

Also we will need to do the same thing for the stop and status.  Please let me know if my assumption for the configuration deployment is correct or not. I will update the dbs/manage script. 

In addition to above for DBS instances, we also have a DBS migration server need to move to k8s. please see dbsmigration/manage.

Yuyi

yuyiguo:
@vkuznet @h4d4 
 I created https://github.com/dmwm/deployment/pull/793  to include stop/status.

vkuznet:
Yuyi,
the DBS system stay apart from other since it uses instance concept. None of the
existing cmsweb services has this feature. As such for every instance there a
single configuration files. So far we keep them together because cmsweb
deployment procedure is based on a single account for single system. In DBS case
we have _dbs account, but we split DBS system into multiple instances.

In k8s I'd like to have DBS instances to have their dedicated pod. This will
allow to scale individual DBS instances dynamically and independently. But
because I follow cmsweb deployment procedure I still use the same deploy/manage
scripts as well as the same manage area.

The tricky part is that we should maintain existing deployment procedure
both in cmsweb and in k8s (at least originally) because we need to perform
migration. Later we can develop own/separate deployment workflow for individual
systems, but so far it is premature to talk about it.

So, what I'll do is to use a trick based on presence of configuration file.
It would be two step procedure for DBS:
- first I'll deploy DBS as is following standard cmsweb deployment procedure,
this is what current k8s install file is doing. It will deploy entire manage
area to the pod, but
- we also have configuration area for k8s where we keep secrets per system.
Instead of single DBS area I'll create separate areas per DBS instance, e.g.
DBS-globalR, DBS-globalW. In each area I'll store proper configuration file
such that during k8s deployment the pod will know which configuration file to
use. In start-up script (for k8s we use individual run.sh script which calls
internally "manage start") I'll check which configuration file was passed around
and it will "wipe out" manage area in a pod and use only that configuration
file.

But in order to make this work I need this check in manage script for presence
of DBS configuration file.

Doing this way we'll have/use the same manage script for both cmsweb deployment
and for k8s one.

Eventually I'll show you how k8s deployment works and it will be more clear.
So far I wanted to have transparent change which will allow to do this trick.

I hope it is clear.
Best,
Valentin.

vkuznet:
close in favor of #793

PR #793, opened 2019-07-12
DBS start/stop/status update for k8s.

yuyiguo:
@vkuznet @h4d4 
Please take a look this PR for DBS in  k8s. This may supersede   https://github.com/dmwm/deployment/pull/792 .

### DIFF
--- a/dbs/manage
+++ b/dbs/manage
@@ -61,6 +61,36 @@ export X509_USER_KEY=$X509_USER_PROXY
 # Necessary to evaluate the configuration
 export DBS3_VERSION
 
+checkCfg()
+{
+
+if [ -f $CFGFILE ]; then
+    echo $CFGFILE 
+fi
+
+if [ -f $CFGFILE_GR ]; then
+    echo $CFGFILE_GR 
+fi
+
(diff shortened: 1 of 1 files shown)


vkuznet:
this looks good to me. Please include it in next cmsweb upgrade.

vkuznet:
Yuyi, upon further investigation I found that these changes will not work. We didn't take into account location of DBS `$CFGFILE`(s).  So we need to use `$CFGDIR/$CFGFILE`, `$CFGDIR/$CFGFILE_GR`, etc. in every `if` statement, but pass only appropriate `$CFGFILE` to `start_dbs` function since it has internally `$CFGDIR/$1`. I suggest to pass to `start_dbs` function full path to cfg file, e.g.

```
start()
{
... --ini=$1
}
start_dbs()
{...
if [ -f $CFGDIR/$CFGFILE ]; then
   start_dbs $CFGDIR/$CFGFILE
fi
...}
etc.
```
please make additional changes accordingly.

vkuznet:
I made another PR to replace this one, please see #794

PR #801, opened 2019-08-08
Add exception for single-user mode to run ServerMonitor w/o sudo

vkuznet:


andrius-k:
@h4d4 Hi Lina,
Could you please also include this PR to the latest CMSWEB release as well?

PR #802, opened 2019-08-14
Split reqmgr2ms into transferor and monitor parts

vkuznet:


amaltaro:
Thanks, Valentin. It sort of works well in general, but the port was clashing with reqmon. I also decided to use a single `interval` configuration for the services polling cycle.

Still have to look into the monitoring.ini file and figure out how to use that.

About the running the services, things apparently run fine, but I get the first started process getting killed. I think it's how the wrapper script deals with services and use a single `pid` file to see the PID of the service (and use it to manage). In short, it tries to manage two services as if there was only one...

I'm creating another PR based on your changes.

amaltaro:
@h4d4 this PR can be closed now

PR #805, opened 2019-08-26
Add process monitor exporters for ms-transferor and ms-monitor services

vkuznet:


amaltaro:
@vkuznet can you also update the other services then such that they follow the same PID file pattern? Here is what I can see in testbed:
```
amaltaro@vocms0731:/data/srv/state $ ls reqmgr2
tmp  reqmgr2.pid
amaltaro@vocms0731:/data/srv/state $ ls workqueue/
cache  globalworkqueue.pid
amaltaro@vocms0731:/data/srv/state $ ls reqmon/
wmstatsserver.pid
amaltaro@vocms0731:/data/srv/state $ ls reqmgr2ms/
tmp  ms-monitor.pid  ms-transferor.pid
amaltaro@vocms0731:/data/srv/state $ ls t0_reqmon/
t0_reqmon.pid
```

PR #811, opened 2019-10-11
Fixing Alignment of Lines on RPC Wheel's Occupancy Histograms

sangyeon83:
Fix the minor bug in RPC Barrel Wheel's Occupancy Histograms. 

- Confusing plots are produced in DQM GUI because of the minor bug in postDrawTH2 part in RPCRenderPlugin.cc. 

- To be specific, drawn lines on Occupancy_Wheel_#_Sector_# does not match with the histograms. 

- Fixing alignment of it was completed and confirmed through DQM GUI local server. 

- Please check it and merge it if it is reliable.

@slezki  @andresib @jhgoh

### DIFF
--- a/dqmgui/style/RPCRenderPlugin.cc
+++ b/dqmgui/style/RPCRenderPlugin.cc
@@ -170,25 +170,25 @@ private:
         TLine line;
         line.SetLineWidth(2);// Draw lines to delimitate the end of the roll
         //rb1in
-        line.DrawLine(91, 0.5, 91, 2.5);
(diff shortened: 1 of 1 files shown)


sangyeon83:
I attach the example of changes. 
Before,
image
After,
image

jhgoh:
This "bug" have been for some years. 
The lines are supposed to give a guideline to show what is maximum number of strips. There was some changes applied to the RPC DQM histogramming part to shift bin center but the layout was not updated to address it.

John for the RPC DPG

PR #812, opened 2019-10-23
changing lemon to collectd command

h4d4:
This PR is to replace lemon by collected command 

Best Regards, Lina.


muhammadimranfarooqi:
merge

PR #823, opened 2019-12-05
Use a single stdout log file

amaltaro:
I have always noticed that we keep 3 logs for CouchDB:
* couch.log (as defined in the local.ini configuration file)
* ###.stdout (standard stdout defined by the manage script)
* ###.stderr (standard stdout defined by the manage script)

where the *only* difference between couch.log and the stdout one, is a timestamp!!

Given that the Couch logs are huge, let's reduce it by half!

### DIFF
--- a/couchdb/manage
+++ b/couchdb/manage
@@ -60,7 +60,7 @@ start()
 {
   cd $STATEDIR
   couchdb -A $CFGDIR/ -a $KEYFILE -b \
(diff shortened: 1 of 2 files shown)


amaltaro:
@todor-ivanov could you please have a look at this one, just to make sure I'm not missing anything.
I quickly tested in central services and the stdout log is no longer there.

PR #825, opened 2019-12-16
Skip couch.log file from the continuous log sync

amaltaro:
@h4d4 @muhammadimranfarooqi as we discussed last Friday, this PR is meant to stop copying the huge couch.log files from the cmsweb backends to vocms055.
From the manpages, that's the correct syntax, but it would be great if one of you could get it testbed and applied to CMSWEB before the Xmas holidays. Thanks

PR #826, opened 2019-12-16
Removing already decommisioned nodes vocms0{127 134} from the list

h4d4:
This PR update the list, removing already decommissioned nodes, those are  vocms0{127 134}


h4d4:
@amaltaro @muhammadimranfarooqi

As we are going to apply this PR: https://github.com/dmwm/deployment/pull/825
I'm also updating the list of nodes, removing already decommissioned nodes, those are vocms0{127 134}

Best Regards, Lina.

PR #828, opened 2020-01-10
Change mount point for the reqmgr2ms

amaltaro:
This change is necessary to mount the MicroService applications in a different path.
With this patch, the MS applications will be mounted against e.g.
`/ms-transferor/data`
instead of 
`/ms-transferor`.

Otherwise a request like `ms-transferor/data/status` uses the `data` REST api, not the `status` one.

### DIFF
--- a/reqmgr2ms/config-monitor.py
+++ b/reqmgr2ms/config-monitor.py
@@ -34,12 +34,12 @@ srv = main.section_("server")
 srv.thread_pool = 30
 main.application = "ms-monitor"
 main.port = 8248  # main application port it listens on
-main.index = 'data' # Configuration requires index attribute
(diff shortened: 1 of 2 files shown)


amaltaro:
Oh well, this change caused the service to duplicate every single record that goes to the logs
```
2020-01-10 19:39:30,412:INFO:MSManager: Total monitor execution time: 0 secs
...
2020-01-10 19:39:30,423:INFO:MSManager: Total monitor execution time: 0 secs
```

We had the same problem in reqmgr2, which was solved with that `log_screen` parameter. It seems not to be working here though...

amaltaro:
Problem was on the application side. This patch is good to go.

PR #822, opened 2019-12-05
Newcouchdb nodes

h4d4:
@amaltaro 
This PR is for include the new couchdb machines, those are vocms074[34] into necessary deployment scripts. The commits were grouped based on the intention of the change, for example:
' Adding vocms074[34] into disable list of services, that should no be installed into this nodes'  98bd4f3 

For below services deployment scripts were not updated, because there is not clear if the service should be installed or no into the new nodes:
acdcserver/deploy
alertscollector/deploy
reqmgr2/deploy
reqmon/deploy
t0_reqmon/deploy
tier0/deploy
wmagent/deploy (I think does not matter because those agents are not deployed into the cluster)
workqueue/deploy

Best Regards, Lina.

PR #831, opened 2020-01-14
Further changes necessary to start using new CouchDB backends

amaltaro:
Follow up of this PR:
https://github.com/dmwm/deployment/pull/822

In short, the first commit contains:
* removal of couchdb service from vocms0741
* and deployment (or not) of services into the two new backends (vocms0743 and vocms0744)
* spread the backup load of couch backups as much as possible

and the second commit does:
* redistribution of the couchdb databases to different backends (which means services are enabled/disabled in different backends)
* thus, new frontend rules (FIXME: `preprod` has some rules from `prod`, such that we can test them before going to production - and rollback)
* fix to the reqmgr2ms/microservices frontend rules
* enable reqmgr2ms on vocms0742

NOTE 1: when merging this PR, do not squash the commits. They address different issues
NOTE 2: before starting services by the end of the deployment (or re-enabling them in the frontends), we need to make sure that the original couch database (the one with meaningful content) has been copied over to the new backend.
NOTE 3: later during the validation phase, I need to create another PR to rollback changes to the preprod backend maps.

@muhammadimranfarooqi @vkuznet you might want to have a first look at it.

For the record, **summary** of the couch databases:
* vocms0740 (workqueue): workqueue and workqueue_inbox
* vocms0742 (reqmgr2, reqmgr2ms and acdcserver): reqmgr_auxiliary, wmdatamining, reqmgr_workload_cache and reqmgr_config_cache
* vocms0743 (reqmon): wmstats, workloadsummary and wmstats_logdb
* vocms0744 (t0_reqmon): tier0_wmstats, t0_request, t0_workloadsummary and t0_logdb

### DIFF
--- a/couchdb/deploy
+++ b/couchdb/deploy
@@ -43,7 +43,7 @@ deploy_couchdb_post()
 {
   (mkcrontab
    case $host in
-     vocms013[689] | vocms073[89] |vocms016[135] | vocms0766 )
+     vocms013[689] | vocms073[89] |vocms016[135] | vocms0741 | vocms0766 )
        disable ;;
      * )
        enable
@@ -62,11 +62,10 @@ deploy_couchdb_post()
          vocms0127 ) tohost=vocms0127 hour=1;;
          vocms0132 ) tohost=vocms0731 hour=1;;
          vocms0731 ) tohost=vocms0132 hour=2;;
-         vocms0740 ) tohost=vocms0731 hour=3;;
-         vocms0740 ) tohost=vocms0743 hour=12;;
-         vocms0742 ) tohost=vocms0744 hour=12;;
-         vocms0743 ) tohost=vocms0740 hour=12;;
-         vocms0744 ) tohost=vocms0742 hour=12;;
+         vocms0740 ) tohost=vocms0742 hour=2;;
+         vocms0742 ) tohost=vocms0743 hour=3;;
+         vocms0743 ) tohost=vocms0744 hour=4;;
+         vocms0744 ) tohost=vocms0740 hour=5;;
                  * ) tohost=;;
        esac
        [ -z "$tohost" ] ||
@@ -76,7 +75,7 @@ deploy_couchdb_post()
 
   # Weekly archive backups to castor
   case $host in
-    vocms012[67] | vocms0132 | vocms0731 | vocms0143 | vocms074[012] | vocms0307 | vocms0318 )
+    vocms0127 | vocms0132 | vocms0731 | vocms074[0234] )
       klist -s # must have afs kerberos token
       (acrontab -l | { fgrep -v -e " $host $project_config/" || true; }
        echo "#30 4 * * 0 $host $project_config/manage archive /castor/cern.ch/cms/archive/cmsweb/backups 'I did read documentation'") | acrontab
--- a/exporters/manage
+++ b/exporters/manage
@@ -123,14 +123,10 @@ start()
               </dev/null 2>&1 | rotatelogs $LOGDIR/das_exporter-%Y%m%d.log 86400 >/dev/null 2>&1 &
         mongodb_exporter -mongodb.uri mongodb://localhost:8230 --web.listen-address ":18230" \
               </dev/null 2>&1 | rotatelogs $LOGDIR/mongodb_exporter-%Y%m%d.log 86400 >/dev/null 2>&1 &
-        wait4proc "couchdb" # check for pattern that couchdb is running
-        couchdb_exporter -couchdb.uri="http://localhost:5984" \
-            -databases="_all_dbs" -databases.views=true -telemetry.address=":15984" \
-              </dev/null 2>&1 | rotatelogs $LOGDIR/couchdb_exporter-%Y%m%d.log 86400 >/dev/null 2>&1 &
       ;;
 
     # CouchDB on production nodes
-    vocms0740 | vocms0743 | vocms0744)
+    vocms0740 | vocms0742 | vocms0743 | vocms0744)
         wait4proc "couchdb" # check for pattern that couchdb is running
         couchdb_exporter -couchdb.uri="http://localhost:5984" \
             -databases="_all_dbs" -databases.views=true -telemetry.address=":15984" \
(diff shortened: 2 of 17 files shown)


amaltaro:
@muhammadimranfarooqi please have a thoroughly review before merging it (and read my comments above)
FYI @vkuznet @todor-ivanov

todor-ivanov:
Thanks @amaltaro

couchdb/deploy
```diff
@@ -62,11 +62,10 @@ deploy_couchdb_post()
          vocms0127 ) tohost=vocms0127 hour=1;;
          vocms0132 ) tohost=vocms0731 hour=1;;
          vocms0731 ) tohost=vocms0132 hour=2;;
-         vocms0740 ) tohost=vocms0731 hour=3;;
-         vocms0740 ) tohost=vocms0743 hour=12;;
-         vocms0742 ) tohost=vocms0744 hour=12;;
-         vocms0743 ) tohost=vocms0740 hour=12;;
-         vocms0744 ) tohost=vocms0742 hour=12;;
+         vocms0740 ) tohost=vocms0742 hour=2;;
```
vkuznet: I still puzzled why tohost does not match hostnames? On line 62 we do have the match, but for last few nodes you added it is not the case. Could you please clarify the logic here.

exporters/manage
```diff
@@ -116,21 +116,17 @@ start()
       ;;
 
     # DAS/MongoDB/CouchDB exporters on production nodes
-    vocms0741 | vocms0742 )
+    vocms0741 )
```
vkuznet: This is wrong since the hostname are for 3 (!!!) services, and it means that das/mongo exporters will not start on vocms0742. Please put back vocms0742 and redo the logic to only exclude couchdb on vocms0742

frontend/backends-preprod.txt
```diff
@@ -3,15 +3,17 @@
 ^/auth/complete/dqm/offline(?:/|$) vocms0738.cern.ch
 ^/auth/complete/dqm/relval(?:/|$) vocms0739.cern.ch
 ^/auth/complete/dqm/(?:dev|offline-test|relval-test)(?:/|$) vocms0731.cern.ch
-^/auth/complete/(?:couchdb2)(?:/|$) vocms0132.cern.ch  :affinity
-^/auth/complete/(?:couchdb|workqueue|wmstats|workloadsummary|alertscollector|tier0_wmstats|t0_workloadsummary|acdcserver)(?:/|$) vocms0731.cern.ch :affinity
-^/auth/complete/(?:c)(?:couchdb|workqueue|wmstats|workloadsummary|alertscollector|tier0_wmstats|t0_workloadsummary|acdcserver)(?:/|$) vocms0731.cern.ch :affinity
+^/auth/complete/(?:c?)(?:couchdb/workqueue|workqueue)(?:/|$) vocms0731.cern.ch :affinity
```
vkuznet: why did you put additional `(?:c?)` here. If you remove bigcouch I think this part should go away.

frontend/backends-preprod.txt
```diff
@@ -3,15 +3,17 @@
...
+^/auth/complete/(?:couchdb/wmstats|wmstats|workloadsummary|wmstats_logdb)(?:/|$) vocms0743.cern.ch :affinity
+^/auth/complete/(?:couchdb/tier0_wmstats|tier0_wmstats|t0_workloadsummary|t0_request|t0_logdb)(?:/|$) vocms0744.cern.ch :affinity
+^/auth/complete/(?:c?)(?:couchdb|acdcserver)(?:/|$) vocms0132.cern.ch :affinity
```
vkuznet: similar argument here, I don't know if we need `(?:c?)` here anymore.

frontend/backends-prod.txt
```diff
@@ -3,15 +3,17 @@
 ^/auth/complete/dqm/offline(?:/|$) vocms0738.cern.ch
 ^/auth/complete/dqm/relval(?:/|$) vocms0739.cern.ch
 ^/auth/complete/dqm/(?:dev|offline-test|relval-test)(?:/|$) vocms0731.cern.ch
-^/auth/complete/(?:couchdb2)(?:/|$) vocms0741.cern.ch :affinity
-^/auth/complete/(?:couchdb/wmstats|wmstats)(?:/|$) vocms0742.cern.ch :affinity
-^/auth/complete/(?:c?)(?:couchdb|workqueue|workloadsummary|alertscollector|analysis_workqueue|analysis_wmstats|tier0_wmstats|t0_workloadsummary|acdcserver)(?:/|$) vocms0740.cern.ch :affinity
+^/auth/complete/(?:c?)(?:couchdb/workqueue|workqueue)(?:/|$) vocms0740.cern.ch :affinity
```
vkuznet: I'm not sure if we need `(?:c?)` anymore.

frontend/backends-prod.txt
```diff
@@ -3,15 +3,17 @@
...
+^/auth/complete/(?:couchdb/wmstats|wmstats|workloadsummary|wmstats_logdb)(?:/|$) vocms0743.cern.ch :affinity
+^/auth/complete/(?:couchdb/tier0_wmstats|tier0_wmstats|t0_workloadsummary|t0_request|t0_logdb)(?:/|$) vocms0744.cern.ch :affinity
+^/auth/complete/(?:c?)(?:couchdb|acdcserver)(?:/|$) vocms0742.cern.ch :affinity
```
vkuznet: I"m not sure if we need `(?:c?)` anymore

couchdb/deploy
amaltaro: In the end it executes an `rsync` command, and if the node is in the same domain, it works with short hostname or fqdn.

exporters/manage
amaltaro: You're right, I'm updating it.

frontend/backends-preprod.txt
amaltaro: I didn't, I simply used the previous command and removed a few databases from the rule. The usage of the extra `c` is sort of a mistery to me (I do think though it's needed for the Futon couch interface).

frontend/backends-prod.txt
amaltaro: same response as above.

amaltaro:
Patch has been updated. Thanks for your comments, Valentin.
I share the same view as you and I'd rather have the explicit nodes to enable our services, however, I think it would cause deployment issues every time an offsite deployment is needed (our private VMs). So either we implement the `variant` feature for every service, or we need to find a way around that.

Anyways, that's out of the scope of this PR and I leave it up to Imran's consideration for the future.

PR #835, opened 2020-01-17
Push couchapps only to specific CouchDB nodes

amaltaro:
With this patch, I changed the model which relies on disabling the service on specific nodes and **enabling** it everywhere else. Now, the service will only be **enabled** on specific nodes, everything else will have it disabled.

It also assumes that if the node hostname does NOT start with `vocms`, then it's a private node and service/couchapps are enabled by default.

In addition to that, we only write to the `stagingarea` in very specific nodes (those running CouchDB as well).

Still to be tested

### DIFF
--- a/acdcserver/deploy
+++ b/acdcserver/deploy
@@ -25,7 +25,7 @@ deploy_acdcserver_sw()
   else :; fi
 }
 
-deploy_acdcserver_post()
+setup_acdcserver_couch()
 {
   # Tell couch to push acdcserver apps on the next restart
   local manage=$project_config/acdcserver_manage
@@ -35,6 +35,24 @@ deploy_acdcserver_post()
     echo "couchapp push $root/current/apps/acdcserver/data/couchapps/GroupUser" \
          "http://localhost:${couch##*:}/acdcserver" >> $root/state/${couch%%:*}/stagingarea/acdcserver
   done
+}
+
+deploy_acdcserver_post()
+{
+  case $host in
+    vocms0117 | vocms0132 | vocms0740 )
+      enable
+      setup_acdcserver_couch
+      ;;
+    * )
+      if [ "$host" != vocms* ]; then
+        enable
+        setup_acdcserver_couch
+      else
+        disable
+      fi
+      ;;
+  esac
 }
 
 deploy_acdcserver_auth()
--- a/reqmgr2/deploy
+++ b/reqmgr2/deploy
@@ -40,33 +40,43 @@ deploy_reqmgr2_sw()
        $root/$cfgversion/config/$project/config.py
 }
 
-deploy_reqmgr2_post()
+setup_reqmgr2_couch()
 {
-  # in practice, general purpose backends and vocms0742 (prod, for couchapps)
-  case $host in
-    vocms013[89] | vocms073[89] | vocms074[0134] )
-      disable;;
-    * )
-      enable;;
-    esac
-
   # ReqMgr2 and old reqmgr specific couchdb stuff
   # Tell couch to pick up reqmgr on the next restart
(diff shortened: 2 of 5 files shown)


amaltaro:
I just tested it on my VM and it works fine. Of course, we need VO nodes (vocmsXXX) to properly test the rest of this PR.
@muhammadimranfarooqi @vkuznet could you please review it? I'm open for suggestions on the check for private nodes BTW.

vkuznet:
Alan, I only wished that you'll wait for our review before the merge. You should be careful in changing now all stuff and consider testing them with k8s setup as well. Usage of individual hostnames is not working in k8s where we don't have explicit hostnames. And, now I need to spend more time of testing if all of your changes will work in k8s.

What should be done instead, is requesting to build docker images and deploying them into k8s with all of your changes related to apps, redirect rules, etc.

amaltaro:
I think this is one of the reasons Imran merged it this afternoon. Because then he can get those tested in the k8s setup as well. Which, AFAIK use a different set of deployment scripts (and frontend rules), so it's unclear to me how it affects the k8s setup.

muhammadimranfarooqi:
@amaltaro @vkuznet 

Yes, I wanted to run HG2002c deployment and then creating docker image for HG2002c release. Thats why I merged it. Currently, I am running deployment for this release.

vkuznet:
In a future please don't merge as we do provide ability to add patches to docker builds. For example, https://github.com/dmwm/CMSKubernetes/blob/master/docker/reqmgr2/install.sh#L57 shows how to do it for given system. Doing this way someone can make a local docker build and then test it before applying the merge which will avoid headache fixing everything afterwards.

And, for the record we deploy services in the same way as we do in VMs, via their deploy script. But, sometime tweaks are necessary to overcome situation when deploy script use explicit names, like this PR, to ensure that k8s deployment will work.

vkuznet:
and, our frontend rules in VM and k8s are the same too. Otherwise we'll never converge and be able to compare two clusters. The only difference is that in k8s we redirect to backend k8s cluster while in VM setup we redirect to specific hostnames.

muhammadimranfarooqi:
@vkuznet 
As we added two new nodes for couchdb and distribute databases of couchdb between those nodes. Will this affect k8s setup after I deploy docker image?

vkuznet:
That's is the question to Alan who is responsible for the service. So far we deployed single CouchDB in k8s which had all databases. In k8s we may either have single setup or we may have multiple pods running CouchDB with independent databases. In former case I think we don't need anything in later we need to write new couchdb manifest files (similar to dbs-XXX ones) where we'll run independent couches (which by itself may require different storages, etc.). I hope Alan should be more aware of this and be involved in designing how service should run.

amaltaro:
@khurtado let me add Kenyi to this discussion. Because he's probably the main person to investigate WMCore services on the k8s land.

vkuznet:
I suggest that DMWM team discuss the architecture how they want to run services on k8s, in particular how many instance of CouchDB to have and how to partition them, which resources are required to each instance, etc. Once we know all details we can adjust k8s manifest files, or better you'll take care of your k8s manifiest files themselves and make appropriate PR to CMSKubernetes repository. If you need guidance from me, feel free to setup a meeting for this specific topic.

amaltaro:
Imram, there is probably a `Revert` button at the bottom of this PR. I don't have enough privileges, so it doesn't show up for me.
If it does for you, can you please go through it (it will create a new pull request reverting exactly the same changes that were proposed in this PR). Once it's done, tag me and I'll have a quick review.

If it does not work, I can create a new PR with the necessary changes in place.

muhammadimranfarooqi:
@amaltaro 

This PR has been reverted as here. https://github.com/dmwm/deployment/pull/838

PR #838, opened 2020-01-20
Revert "Push couchapps only to specific CouchDB nodes"

muhammadimranfarooqi:
Reverts dmwm/deployment#835


muhammadimranfarooqi:
As discussed by Alan, previous PR did not work for having only databases in specific nodes. So reverting this on the request of Alan.

Best

PR #839, opened 2020-01-23
Adding hostnames in logs of all manage scripts

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

Could you please review the PR. I have added hostname in logs so that they can be distinguished in the k8s logs. 

BR

acdcserver/acdcserver_manage
```diff
@@ -27,7 +27,7 @@ export WMCORE_ROOT=$ACDCSERVER_ROOT
 export YUI_ROOT
 export ACDC_CLI="${ACDCSERVER_ROOT}/bin/acdcserver-tools"
 export WMAGENT_CONFIG=${CFGFILE}
-logcmd="rotatelogs $LOGDIR/acdcserver-%Y%m%d.log 86400 2>&1"
+logcmd="rotatelogs $LOGDIR/acdcserver-`hostname`-%Y%m%d.log 86400 2>&1"
```
vkuznet: Please change `hostname` to `hostname -s` as we don't want .cern.ch in a log file name. This should apply to all changes in this PR.

vkuznet: I would also suggest to change file name to this format:
<service>-<date>-<host>.log
Doing this way it would be easier to sort files by a services/date. For instance if I want to cat all today's files for services dbs, I'll do `cat dbs-20200123-*.log`. Moreover the pods inherit service names too, therefore using convention you have will produce `dbs-dbs-global-r-20200123.log` which is kind of weird, while separating service from host name with a date in a middle will probably easier to understand.

muhammadimranfarooqi: @vkuznet 
I did not understand this filename format example. Could you please elaborate more. 

Thanks

amaltaro:
Imran, can you or Valentin please clarify why we need to change the log file name? Is it due to the problem Stefano reported in yesterday's meeting? Can you please elaborate what happens to those logs (maybe there is another alternative then renaming every single log file).

acdcserver/acdcserver_manage
vkuznet: The format
```
acdcserver-`hostname`-%Y%m%d.log
```
will produce this file name
```
acdcserver-somehost.cern.ch-20200123.log
```

What I'm saying that we should use short host name (without domain) and change order in filename, i.e. this format
```
acdcserver-%Y%m%d-`hostname -s`.log
```
will produce
```
acdcserver-20200123-somehost.log
```
(here I used `somehost.cern.ch` as a hostname example and `somehost` as short name of hostname).

muhammadimranfarooqi: @vkuznet 

Ok. I understood now. I'll change according to your suggestions.

Thanks.

muhammadimranfarooqi:
> Imran, can you or Valentin please clarify why we need to change the log file name? Is it due to the problem Stefano reported in yesterday's meeting? Can you please elaborate what happens to those logs (maybe there is another alternative then renaming every single log file).

Yes. It is related to Stefano's comments in yesterday's meeting. By adding hostname with logs, it would be able to identify the pod from which logs are generating.

vkuznet:
Alan, my fear that if we run multiple pods and all of them will have single file name we may have racing/collision problem within single log file from multiple pods. Therefore it is better to have separate log files within each pod but let them reside in common shared cephfs area. The only way to achieve that is to add proper hostname (pod) to the log file name.

amaltaro:
Ok, so a CEPH volume will be mounted on every single node, the container will use a bind mount and write all its logs to that CEPH volume (which is also used by all the other pods).
Yes, adding a short hostname to the log will do the trick then (but I'd avoid the double dash if possible).

vkuznet:
yes, cephfs volume for logs is mounted in every pod. The pod names (determined by `hostname -s`) have this structure (I take crabserver as an example): `crabcache-55b67f4b76-bxz4g`. The hashes are assigned by k8s, and there is no double dashes, i.e. the final log name will be (in this case): `crabserver-20200123-crabcache-55b67f4b76-bxz4g.log`

muhammadimranfarooqi:
@vkuznet , @amaltaro 

Could you please review this PR now.

Thanks

belforte:
looks good to me, thanks.

amaltaro:
Changes look good to me as well. I'll have to update the log name of the cherrypy threads as well (both in the deployment and in WMCore)

muhammadimranfarooqi:
> Changes look good to me as well. I'll have to update the log name of the cherrypy threads as well (both in the deployment and in WMCore)

Do you want me to change the name? I can do with this PR.

amaltaro:
Thanks, Imran. It would be great if you could have a look at the config files for a few services that also define a log file name (e.g. reqmgr2, workqueue, reqmgr2ms, reqmon, and maybe a few others that don't come to my mind at the moment)

belforte:
Now that you make me think about it, there are these CRAB-*log  files as well,
which I do not see on vocms0750
```
belforte@vocms055/crabserver> pwd
/build/srv-logs/vocms0163/crabserver
belforte@vocms055/crabserver> ls
CRAB-20200115.log	 crabserver-20200117.log  crabserver-20200120.log  crabserver-20200123.log
CRAB-20200119.log	 crabserver-20200118.log  crabserver-20200121.log
crabserver-20200116.log  crabserver-20200119.log  crabserver-20200122.log
belforte@vocms055/crabserver> 

```

should I change something at my end ?
(no, I do not know why CRAB Server writes two log files... it is not my code, and documentation died young)

vkuznet:
@muhammadimranfarooqi changes looks good to me

vkuznet:
@belforte , Stefano, I suggest that you inspect your service and identify who produces CRAB files. We should not be involved. Then inspect your crab pods and see what they have written in logs area. Then inspect your logs in vocms055 and identify which content is in CRAB files, etc. Then report what is missing.

muhammadimranfarooqi:
> Thanks, Imran. It would be great if you could have a look at the config files for a few services that also define a log file name (e.g. reqmgr2, workqueue, reqmgr2ms, reqmon, and maybe a few others that don't come to my mind at the moment)

@amaltaro 
Alan its done. Could you please review this.

amaltaro:
Thanks, Imran. Changes look good to me. Please merge this one before the other 2 that are removing some deployment scripts.

PR #848, opened 2020-02-07
Move ReqMon CherryPy threads from vocms0740 to vocms0743

amaltaro:
Imran, Todor,

this configuration detail was missed when I was working on the deployment changes required for the two new couchdb instances: https://github.com/dmwm/deployment/pull/831

Besides merging it for the March release, we need to update the reqmon configuration on both vocms0740 and vocms0743 and restart the `reqmon` service.


amaltaro:
Imran and I have applied this change to both backends and the service is running properly from vocms0743 now.

PR #849, opened 2020-02-07
GEM online DQM updates - full GE1/1

quark2:
#### PR description:
This PR updates and extends the part of onlineDQM for GEM. The details are following: 

 - The existing parts are fixed to match the VFAT version which is used in CMSSW currently
 - Summary plot is added
 - Histograms which displays the status of devices (AMCs, GEBs, VFATs) by time are added

Please see [this](https://indico.cern.ch/event/870366/contributions/3670607/attachments/1983858/3304860/20200207_Byeonghak_Ko_GEM_OnlineDQM.pdf) to check more details

@jshlee @watson-ij


andrius-k:
Hi @quark2,

I tried to apply this PR when deploying the playback DQM GUI and I got this error:

```
/data/srv/current/config/dqmgui/style/GEMRenderPlugin_SummaryChamber.cc: In constructor 'SummaryChamber::SummaryChamber()':
/data/srv/current/config/dqmgui/style/GEMRenderPlugin_SummaryChamber.cc:64:3: error: 'sort' is not a member of 'std'
   std::sort(listChPre.begin(), listChPre.end(), lambdaChamber);
   ^~~
make: *** [/data/srv/state/dqmgui/online/render/GEMRenderPlugin_SummaryChamber.o] Error 1
```

Could you please take a look at this. You can find the instructions on how to develop the GUI locally here: https://twiki.cern.ch/twiki/bin/view/CMS/DQMGuiForUsers#Reference_transcript

Regards,
Andrius

andrius-k:
Hi @quark2,

Actually I tried deploying the GUI without your this PR merged and I'm still getting the same error. I will continue debugging on Monday.

Regards,
Andrius

quark2:
Hi @andrius-k, 

It's strange. The problematic line is not in the original one.

Btw, I've fixed that. It has been tested by the transcript ([1]). I found that I used other release and gcc than that in the transcript...


Best regards, 
Byeonghak Ko


[1] https://twiki.cern.ch/twiki/bin/view/CMS/DQMGuiForUsers#Reference_transcript

andrius-k:
Hi @quark2,
Indeed very strange. When I tried deploying today everything worked totally fine. Anyway, this PR is already applied to the playback DQM GUI (https://cmsweb.cern.ch/dqm/online-playback/) so you can check it out and let me know if everything is fine.
Regards,
Andrius

quark2:
Hi @andrius-k, 

Ahh, okay. Actually the error is also questionable. std::sort is in gcc/g++ long time ago, but the compiler denied. I think it's weird.

Anyway, it's not a problem now and I think there is no problem, also in the playback. Thanks!


Best regards, 
Byeonghak Ko

andrius-k:
Hi @quark2,
Cool! Does that mean that we can deploy the same version of the GUI to the production system for the upcoming MWGR#1 as well?
Regards,
Andrius

quark2:
Hi, @andrius-k, 

If you mean the version used in the transcript you gave, then I think it will work well, which is tested in private. But I'm not sure which version you meant.

Best regards, 
Byeonghak Ko

andrius-k:
Hi @quark2,
I mean the version that I deployed to the playback today that has this very PR included.
Regards,
Andrius

quark2:
Hi @andrius-k, 

GUI looks fine. But I found one problem; the summary plot isn't generated. It looks like it's because validation part wasn't executed.

Best regards, 
Byeonghak Ko

andrius-k:
Hi @quark2,
We are replaying the run 334614 in the playback system. Are you expecting to see a summary plot being filled in with this data?

quark2:
Hi @andrius-k, 

I expect there will be the summary plot no matter if GEM data are taken or not. Ah, btw, I made a mistake on my word. I said "validation part", but it should be "harvesting part". Is it run in the playback system?

Best regards, 
Byeonghak Ko

andrius-k:
Hi @quark2,

Indeed that is the reason why summary plot is not filled. In the Online, harvesters are called on `endLumi` and not in `endJob` like in the Offline.  As far as I can see, your harvesting logic is in `endRun`: https://github.com/cms-sw/cmssw/blob/master/DQM/GEM/plugins/GEMDQMHarvester.cc#L52

So you should extract the logic in a different method and call it from both `endLumi` (for the Online) and `endJob` (for the Offline) if you want the same harvesting logic to be applied in both cases.

Keep in mind that for the Offline it's `endJob` and not `endRun` like you have now.

Regards,
Andrius

quark2:
Hi @andrius-k, 

Ahhh, thanks! Then should I fix it immediately, to make it run on `endLumi`...? 

Best regards, 
Byeonghak Ko

andrius-k:
Hi @quark2,
Yes, please provide a PR and I will deploy it to the playback system.  Make sure to provide a backport to 11_0_0 as well.
Regards,
Andrius

quark2:
Hi @andrius-k, 

Could you check if this fix makes sense? I changed `endRun` to `dqmEndLuminosityBlock`. I cannot use `endLumi`, but I think it could be another way, and I checked it works.
https://github.com/quark2/cmssw/blob/GEMOnlineDQMBugFix_11_0_X/DQM/GEM/plugins/GEMDQMHarvester.cc

If it's "logically" fine, I'll PR this immediately. Sorry about this mistake...

Best regards, 
Byeonghak Ko

quark2:
Hi @andrius-k, 

Um, I think it's better to PR first. Here is the PR.
https://github.com/cms-sw/cmssw/pull/28977

Best regrads, 
Byeonghak Ko

PR #854, opened 2020-02-14
Add runregistry frontend rule

fabioespinosa:
Chose arbitrarily port 9500 for my ingress controller, please let me know if this port is feasible.

### DIFF
--- a/frontend/app_runregistry_nossl.conf
+++ b/frontend/app_runregistry_nossl.conf
@@ -1,0 +1,1 @@
+RewriteRule ^(/runregistry(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUERY} [R=301,NE,L]
--- a/frontend/app_runregistry_ssl.conf
+++ b/frontend/app_runregistry_ssl.conf
@@ -1,0 +1,2 @@
+RewriteRule ^(/runregistry(/.)?)$ /auth/verify${escape:$1} [QSA,PT,E=AUTH_SPEC:limited-proxy;cert]|
+RewriteRule ^/auth/complete(/runregistry(/.)?)$ http://%{ENV:BACKEND}:9500${escape:$1} [QSA,P,L,NE]|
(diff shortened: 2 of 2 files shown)


vkuznet:
Fabio, check this page https://cms-http-group.web.cern.ch/cms-http-group/activity.html which lists our port allocations. The services still use 8XYY ports. Please use next vailable slt, i.e. 8400-8409 range and use it for your app.

fabioespinosa:
Thank you Valentin, I checked the webpage and updated this PR with port 8350.

I also noticed DQM GUI is listed twice (8040-8049 and 8060-8089) is this desired?

Also submitted merge request on gitlab to update the webpage you sent me: https://gitlab.cern.ch/cms-http-group/doc/merge_requests/1/commits

vkuznet:
ok, that's fine. Please also add frontend/app_runregistry_nossl.conf redirect rule.

vkuznet:
regarding dqm gui, I don't know whole legacy, most likely they added new dev/relval instances on different port range.

fabioespinosa:
Ok, added rule for app_runregistry_nossl.conf, based on other files I see that there is no port specified. I wonder how this will map to the ingress controller. I followed the other examples.

Let me know if I need to change it.

vkuznet:
ok, rules looks good to me. The nossl rules just redirect request from http to https, that's why there is no port number. The ssl rule redirect from FE url to backend one where you have your service running on that port. These rules have nothing to do with ingress in k8s per se, because we apply them to our apache which can be run in VM cluster or in k8s one. But in k8s manifest files we use whatever port you use for your service, in that case you choose to use port 8350.

fabioespinosa:
I see, thanks for the explanation, I will proceed trying to deploy runregistry in cmsweb-test6.cern.ch (since it has the highest kubernetes version).

fabioespinosa:
Hi Valentin, anything missing from my side in this PR?

vkuznet:
Fabio, everything is fine, but I don't have rights to merge this PR, it is up to @muhammadimranfarooqi Muhammad.

PR #875, opened 2020-03-23
Use CouchDBCleanup thread to clean transfer documents too

amaltaro:
Added a new variable for one of the cherrypy threads:
```
couchCleanup.auxCleanDuration
```
such that that task runs every 12h.


amaltaro:
This PR is also good to go, but please merge it only **after** (once there are no conflicts found):
https://github.com/dmwm/deployment/pull/876

PR #876, opened 2020-03-23
Move parent dataset lock thread from reqmon to reqmgr

goughes:
Moving the thread to lock parent datasets from reqmon to reqmgr. Running this in reqmon required running the same thread on multiple backends, which could create issues. Moved to reqmgr2 where a single instance can manage parent dataset locking.

Needed to support changes from this PR:
https://github.com/dmwm/WMCore/pull/9567


amaltaro:
@muhammadimranfarooqi this pull request must be merged **before**:
https://github.com/dmwm/deployment/pull/875

BTW, can you please also update the CMS HTTP documentation by adding this new "ReqMgr2 BuildParentLock thread" to vocms0766 and vocms0731, under:

PR #882, opened 2020-04-04
Deploy mongodb with reqmgr2ms

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/9605

We are going to start working on a third microservice to be provided by this reqmgr2ms package. It will depend on MongoDB, thus deploy it as well.

I noticed that the DAS deploy script also deploys the `admin` package, I'm not sure what that actually deploys, so I'll keep it out for now.


amaltaro:
This is good to go.

PR #883, opened 2020-04-06
Fix order of couchdb backend rules

amaltaro:
Imran, as we discussed over the weekend, this is the reason why wmstats documents are stored in vocms0132 instead of vocms0731.

These rules are evaluated in a top to bottom order, which means vocms0132 is getting documents for:
workqueue
workqueue_inbox
wmstats
workloadsummary
wmstats_logdb
and those should be going to vocms0731.

This will require manual intervention of the databases, moving (or if we need, replicating) databases from one node to another.


amaltaro:
This is good to go.

PR #885, opened 2020-04-08
pass rucio_host and rucio_auth to the agent configuration

amaltaro:
Also pass the `rucio_host` and `rucio_auth` to the wmagent initialization.


amaltaro:
This is good to go.

PR #890, opened 2020-04-14
Introducing the msOutput configuration.

todor-ivanov:
DO NOT merge yet! 
WORK IN PROGRESS

This is an initial request, needed for the service deployment during the development phase.


amaltaro:
@todor-ivanov I think the reason our https requests don't go through is because we have missed the frontend rewrite rules. I made a fix here: https://github.com/dmwm/deployment/pull/891

PR #892, opened 2020-04-24
Fix missing backend redirection rules for ms-output.

todor-ivanov:
@muhammadimranfarooqi This is a bug fix for the deployment scripts for MSOutput. Please see:
https://github.com/dmwm/WMCore/issues/9661

### DIFF
--- a/exporters/manage
+++ b/exporters/manage
@@ -113,6 +113,8 @@ start()
               </dev/null 2>&1 | rotatelogs $LOGDIR/ms-transferor_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
         process_monitor.sh /data/srv/state/reqmgr2ms/ms-monitor.pid ms-monitor ":18248" 15 \
               </dev/null 2>&1 | rotatelogs $LOGDIR/ms-monitor_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
+        process_monitor.sh /data/srv/state/reqmgr2ms/ms-output.pid ms-monitor ":18245" 15 \
+              </dev/null 2>&1 | rotatelogs $LOGDIR/ms-output_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
       ;;
 
     # DAS/MongoDB exporters on vocms0741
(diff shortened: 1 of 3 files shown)


exporters/manage
```diff
@@ -113,6 +113,8 @@ start()
               </dev/null 2>&1 | rotatelogs $LOGDIR/ms-transferor_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
         process_monitor.sh /data/srv/state/reqmgr2ms/ms-monitor.pid ms-monitor ":18248" 15 \
               </dev/null 2>&1 | rotatelogs $LOGDIR/ms-monitor_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
+        process_monitor.sh /data/srv/state/reqmgr2ms/ms-output.pid ms-monitor ":18245" 15 \
```
amaltaro, line 116: @muhammadimranfarooqi do you need to open this port somewhere?

muhammadimranfarooqi: @amaltaro 

I don't know. I can test it during deployment procedure.

vkuznet:
The port for any exporter should be accessible within the CERN network, we run the Prometheus server in CMS k8s cluster and it scrapes metrics from exporters. So, if you want these metrics visible and build dashboard the port should be open to cms-monitoring.cern.ch nodes.

But, if, for some reason, you cannot open the port but still want to have metrics, we provide CLI solution, i.e. run local Prometheus and use terminal-based dashboards. In this case, there is no need to open a port.

amaltaro:
We probably want to use the same prometheus monitoring mechanism used by other services,. Thus, that port should be opened such that we automatically get this data in MonIT.

Just one clarification though, I assume that port has to be opened on the backend node (where the service runs AND where the exporter will run), right? Meaning, the prometheus nodes don't actually talk to the backends through the frontends, but they go straight to the backends. Have I missed anything?

vkuznet:
Yes, the cms-monitoring.cern.ch (where Prometheus runs) talks directly to backend nodes where exporters run. Therefore we open up port ONLY on CERN network between the nodes which run exporters and cms-monitoring.cern.ch. 

You should also understand that it is not automatic as we need to add node:port to Prometheus config once the port is open. Once we change Prometheus configuration and add new node:port we restart Prometheus and the reset is automatic, i.e. our Prometheus metrics are visible in CERN MONIT Grafana. This is how we monitor all services/nodes via exporters.

amaltaro:
Perfect, thanks for the clarification Valentin.
@muhammadimranfarooqi @todor-ivanov this PR needs further changes. I'll cherry-pick your commit and start a new PR.

todor-ivanov:
Thanks @amaltaro

amaltaro:
Superseeded by https://github.com/dmwm/deployment/pull/894

muhammadimranfarooqi:
@amaltaro 
Can I close this PR?

PR #894, opened 2020-04-28
Add backend rules for ms-output; fix reqmgr2ms exporters

amaltaro:
In addition to adding the preprod and prod rules for `ms-output` service; this PR also moves the `reqmgr2ms` exporters to the correct nodes:
* for production: vocms0742
* for preprod: vocms0731

Imran, I'll update the gitlab ticket with further information about those ports.

@vkuznet Valentin, as you can see, reqmgr2ms exporters are enabled for both pre-prod backends, even though we only use it in vocms0731. My question to you is, do you prefer to have a bunch duplicated code in there (but only enable things in the correct backend); or stick to what is provided in this PR, which avoids code duplication in the exporters/manage script, but runs process_monitor.sh for PIDs that won't actually exist? Please let me know and I can adapt this PR.


vkuznet:
Alan, I think we should do proper clean-up re-factoring of manage script, i.e. the exporters should be started only on relevant nodes.

The issue here is that originally we have a small list of exporters and their grouping was relevant. Now, this list growth, and it would be better to reorganize the manage script. Please make proper adjustments.

amaltaro:
Valentin, I can make further changes related to this pull request such that we don't run reqmgr2ms exporters where there is no reqmgr2ms.

About the refactoring, it's probably better to deal with that in a different pull request. I would also probably need some input from you and Imran to confirm which exporters are executed in every single node of the CMSWEB cluster(s).

amaltaro:
@vkuznet @muhammadimranfarooqi I actually went ahead and did most of the refactoring here, such as:
* only enable exporters on the correct nodes
* avoid duplicate code

Valentin/Imran, please carefully review those changes to make sure I did not miss anything. I'm happy to make further fixes in another PR, if needed.

Imran, these changes are a bit harder for me to test. So, if you want to proceed with testbed this morning, I'm all for it.

muhammadimranfarooqi:
@vkuznet @amaltaro 

Is this PR ready to go in ?

exporters/manage
```diff
@@ -71,171 +71,172 @@ check()
   fi
 }
 
+apache_monitors()
+{
+    # start apache_exporter for frontend
+    wait4proc "httpd" # check for pattern that httpd is running
+    apache_exporter --telemetry.address ":18443" \
+          </dev/null 2>&1 | rotatelogs $LOGDIR/apache_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
+    node_exporter --web.listen-address=":19000" \
```
vkuznet, line 80: I suggest putting node_exporter into individual node_monitors() function. The node exporter is generic one which can monitor any Linux nodes, therefore we may include it into any cmsweb FE/BE nodes if we want to.

amaltaro:
Thanks for the review, Valentin. I have pushed those changes in in the 3rd commit.
Imran, please let me know if you want me to squash those commits from my side.

Other than that, this PR is good to go from my side.

PR #895, opened 2020-04-29
Frontend rules for k8s testbed cluster

muhammadimranfarooqi:


vkuznet:
I suggest that you name this file as backends-k8s.txt and then you may modify frontend install.sh script to use it. Doing this way we can separate all deployments.

muhammadimranfarooqi:
@vkuznet 

I have renamed as per your suggestion. I checked phedex works perfect. However, for couchdb I get authorization failed error. 
https://cmsweb-k8s-testbed.cern.ch/couchdb/_utils/
{"error":"unauthorized","reason":"Authorisation failed"}

For phedex nodes, i needed to open firewall for k8s frontend. Same I did for couchdb node . 
https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsweb/-/blob/cmswebdev/code/manifests/firewall.pp#L125

Any idea what could be reason for this? Thanks

PR #897, opened 2020-05-04
Adopt WMArchive scripts to use both Go and Python based deployment

vkuznet:
This PR  provides a way to perform transparent migration from Python-based service to Go based one.


vkuznet:
@muhammadimranfarooqi , this PR can be safely merged as it only affects changes internal to WMArchive.

PR #898, opened 2020-05-05
Do not monitor couch views; monitor specific dbs in each backend

amaltaro:
This PR contains 4 changes:
* given that `node_monitors` is already started in every single backend, I got rid of the line duplication
* check status of `node_exporter` on every single backend (status does not consider that exporter at this very moment)
* disabled monitoring of couchdb views
* make specific monitoring of couchdb databases; only monitor databases under use on each of those backends

If this change looks good. I would actually like to have this being pushed to all the couchdb backends, especially in production (the view monitoring is too heavy on already very loaded services).

@muhammadimranfarooqi @vkuznet could you please review this PR?

### DIFF
--- a/exporters/manage
+++ b/exporters/manage
@@ -153,7 +153,7 @@ couchdb_monitors()
     # start couchdb exporters
     wait4proc "couchdb" # check for pattern that couchdb is running
     couchdb_exporter -couchdb.uri="http://localhost:5984" -logtostderr=true \
-        -databases="_all_dbs" -databases.views=true -telemetry.address=":15984" \
+        -databases=$1 -databases.views=false -telemetry.address=":15984" \
         </dev/null 2>&1 | rotatelogs $LOGDIR/couchdb_exporter-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
 }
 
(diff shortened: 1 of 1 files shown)


exporters/manage
```diff
@@ -153,7 +153,7 @@ couchdb_monitors()
     # start couchdb exporters
     wait4proc "couchdb" # check for pattern that couchdb is running
     couchdb_exporter -couchdb.uri="http://localhost:5984" -logtostderr=true \
-        -databases="_all_dbs" -databases.views=true -telemetry.address=":15984" \
+        -databases=$1 -databases.views=false -telemetry.address=":15984" \
```
vkuznet, line 156: Alan, I don't know for sure if double quotes are required when you pass through input argument. You should check that by yourself and verify either case. So it may be that we need `-databases="$1"`.

amaltaro:
According to the documentation, there is no need for the double quotes:
https://github.com/gesellix/couchdb-prometheus-exporter#database-disk-usage-stats

nonetheless, I did run a test (in vocms0731 :D) and it successfully stopped/started things.

amaltaro:
Imran, any chance we could apply it to the testbed nodes tonight (and at least to vocms0740)? Such that we can reduce the load on couchdb. It's just a matter of:
* replacing the manage script (keeping same ownership/permissions)
* stop/restart exporters

amaltaro:
As discussed in that private thread, these changes have been applied to both vocms0731 and vocms0740, for now.

PR #901, opened 2020-05-13
New frontend redirection rules for preprod and prod for k8s cluster

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

I have a question. In this rule ^/auth/complete/httpgo(?:/|$) cmsweb-srv.cern.ch , the hostname should be of frontend or backend? As httpgo is deployed in both clusters. 

What about httpsgo? Should I include something like ^/auth/complete/httpsgo(?:/|$) cmsweb-srv.cern.ch as well?

Thanks

vkuznet:
The httpgo and httpsgo are simple services which are doing nothing except printing HTTP headers. We use them mostly for testing purposes. For instance, I can use httpgo service on FE just to test how cinder volume works, etc.

Said that, the rules we need are only for BE, so we need /auth/complete/httpgo points to BE, the same for httpsgo. On FE we should not expose anything for security reason, but we may run httpgo/httpsgo if we want to test something there.

muhammadimranfarooqi:
Thanks Valentin

PR #907, opened 2020-05-28
Remove old frontend rules for bigcouch setup

amaltaro:
This PR reverts all the special frontend setup put in place for dealing with both CouchDB and BigCouch in CMSWEB, see:
https://github.com/dmwm/deployment/commit/87f36dde6cfe21e5b0521e548213a229cc8df2e3

Given that BigCouch is no longer in our stack, I decided to clean those rules as much as possible.
@muhammadimranfarooqi @vkuznet can you please review? Once this gets merged, I have further couchdb rules to update.


frontend/app_couch_ssl.conf
```diff
@@ -1,22 +1,22 @@
...
-RewriteRule ^/auth/complete/couchdb2?/tier0_wmstats/_design/WMStats/_view/requestHistory(/.*)?$ http://%{ENV:BACKEND}:5984/tier0_wmstats/_design/WMStatsErl/_view/requestHistory${escape:$1} [QSA,P,L,NE]
-RewriteRule ^/auth/complete/couchdb2?/tier0_wmstats/_design/WMStats/_view/jobsByStatusWorkflow(/.*)?$ http://%{ENV:BACKEND}:5984/tier0_wmstats/_design/WMStatsErl/_view/jobsByStatusWorkflow${escape:$1} [QSA,P,L,NE]
-RewriteRule ^/auth/complete/couchdb2?/tier0_wmstats/_design/WMStats/_view/allWorkflows(/.*)?$ http://%{ENV:BACKEND}:5984/tier0_wmstats/_design/WMStatsErl/_view/allWorkflows${escape:$1} [QSA,P,L,NE]
-RewriteRule ^/auth/complete/couchdb2?/tier0_wmstats/_design/WMStats/_view/time(/.*)?$ http://%{ENV:BACKEND}:5984/tier0_wmstats/_design/WMStatsErl/_view/time${escape:$1} [QSA,P,L,NE]
-RewriteRule ^/auth/complete/couchdb2?(/.*)?$ http://%{ENV:BACKEND}:5984${escape:$1} [QSA,P,L,NE]
+RewriteRule ^(/couchdb(/.*)?)$ /auth/verify${escape:$1} [QSA,PT,E=AUTH_SPEC:cert;host]
+RewriteRule ^/auth/complete/couchdb/wmstats/_design/WMStats/_view/agentInfo(/.*)?$ http://%{ENV:BACKEND}:5984/wmstats/_design/WMStatsErl/_view/agentInfo${escape:$1} [QSA,P,L,NE]
```
vkuznet, line 2: Alan, in this case, I don't know if we really need rules for all paths without couchdb one. Instead, what we need is explicit rules for `/wmstats`, etc. which we use in HTTP calls. For instance, we access wmstats as `https://cmsweb.cern.ch/wmstats` rather than `https://cmsweb.cern.ch/couchdb/wmstats`. The same will be applied to workqueue, etc. I think we need rules like:
```
# couchdb rules
RewriteRule ^(/couchdb(/.*)?)$ /auth/verify${escape:$1} [QSA,PT,E=AUTH_SPEC:cert;host]
RewriteRule ^/auth/complete/couchdb(/.*)?$ http://%{ENV:BACKEND}:5984${escape:$1} [QSA,P,L,NE]

# wmstats rules
RewriteRule ^(/wmstats(/.*)?)$ /auth/verify${escape:$1} [QSA,PT,E=AUTH_SPEC:cert;host]
RewriteRule ^/auth/complete/wmstats(/.*)?$ http://%{ENV:BACKEND}:5984${escape:$1} [QSA,P,L,NE]

etc....
```

Could you please re-evalute and test this?

If this is the case, we should probably add new files for each individual service, e.g. keep couchdb rules in frontend/app_couch_*ssl.conf, while keep wmstats in its own file frontend/app_wmstats_*ssl.conf This will become much more clear since we will keep all individual rules for each individual services we have.

amaltaro: Valentin, we do access couchdb in the form of:
```
https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStats/_view/requestHistory
```
which then sort of makes sense to have these rules here.

amaltaro:
All the other backends-*.txt files are up-to-date and no further changes are required.
I'd rather go with a safe change here, even if that does not give us an optimal cleanup.

frontend/app_couch_ssl.conf
vkuznet: ok, then

vkuznet:
the content of PR looks good to me.

amaltaro:
Thanks, I have ran some basic tests and everything looks ok too.

PR #909, opened 2020-06-03
Adding new R9 plots to ECAL

alejands:
Since some sections of the current ECAL DQM code were written originally, there have been some minor changes in how variables important to the monitoring are commonly understood by ECAL experts. Specifically, there are a few ways of defining R9, all of which might give slightly different values for the variable -- whether to use the zero-suppressed energy, whether to use the raw values for the energy, and so on. And so the current plots are sometimes a little difficult to interpret -- for example, there was an inconsistency noted in the R9 1D distributions between UL and EOY rereco for some samples, and we are investigating whether this has to do with the old definition. This PR adds these new R9 plots to the Ecal workspace.


andrius-k:
Hi @alejands,
If you would like this PR to be deployed, please indicate so here: https://gitlab.cern.ch/cms-http-group/doc/issues/218
Regards,
Andrius

PR #913, opened 2020-06-04
Check presence of config before starting the service

vkuznet:
This change is required to separate services in k8s, i.e. if service config is present the service will be started, otherwise, no action will be done.

### DIFF
--- a/reqmgr2ms/manage
+++ b/reqmgr2ms/manage
@@ -62,7 +62,7 @@ fi
 # Start service conditionally on crond restart.
 sysboot()
 {
-if [ -f $CFGDIR/$CFGFILETR ]; then
+if [ -f $CFGFILETR ]; then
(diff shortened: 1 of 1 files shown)


vkuznet:
@amaltaro , @muhammadimranfarooqi  please review these changes. As I wrote in email they are required to start ms services independently in k8s environment

reqmgr2ms/manage
```diff
@@ -62,38 +62,61 @@ fi
 # Start service conditionally on crond restart.
 sysboot()
 {
+if [ -f $CFGDIR/$CFGFILETR ]; then
```
amaltaro, line 65:
```suggestion
if [ -f $CFGFILETR ]; then
```

vkuznet: sorry, I missed it, now it is fixed, please review again.

amaltaro:
This PR is missing to close an `if` block. I tried to push changes into this branch, but I have no permissions. So I created a new PR here:
https://github.com/dmwm/deployment/pull/916

this one can be closed.

PR #915, opened 2020-06-09
Wait for mongoDB service to start before reqmgr2ms.

todor-ivanov:
This change is supposed resolve at startup the new service dependency from mongoDB introduced by the MSOutput component.


vkuznet:
+1

PR #916, opened 2020-06-09
Start reqmgr2ms services according to the config files available

amaltaro:
Superseeds https://github.com/dmwm/deployment/pull/913

Following the separation of reqmgr2ms services into different PODs, see:
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/6

we noticed that every POD was actually running all 3 services. 
In adittion to these changes applied at when creating a docker container:
https://github.com/dmwm/CMSKubernetes/commit/86f83f3fbca3a4571e54bec2d8f19dd3a18d232a
https://github.com/dmwm/CMSKubernetes/commit/5f195f62de26e5979401ac06dc63feba226f151c

Valentin also suggests to implement the same approach as done by DBS, in order to start services based on the configuration files locally available. Which is what is provided by this PR


amaltaro:
It's been tested and it's good to go

vkuznet:
+1

PR #917, opened 2020-06-12
Move wait4db to section of output ms since it is not used by other ms services

vkuznet:


vkuznet:
@amaltaro , @todor-ivanov , @muhammadimranfarooqi  please review.
I found that you call wait4db for all MS services, while from your k8s configuration it seems it only required for ms output microservice.

If you leave it for all then you should start MongoDB for all of them in k8s and in VM.

amaltaro:
Changes look good to me. Thanks Valentin.
I suggest we wait for Kenyi to test it though. If it's too late for him, I'll test these tomorrow.

khurtado:
Just tested the changes and they are good. +1 for merging

PR #918, opened 2020-06-15
Adding logarchive script for k8s logs mounted at vocms0750 node

muhammadimranfarooqi:


muhammadimranfarooqi:
@vkuznet 

I tested this script on vocms0750 and it worked perfectly. So, this script can be put directly in the cron job of vocms0750. Nothing needs to be changed on the k8s or in deployment scripts.

vkuznet:
ok

PR #920, opened 2020-06-15
Removing view_instances_k8s from deploy script

muhammadimranfarooqi:


muhammadimranfarooqi:
@yuyiguo 

Also removing k8s view instances from deploy script.

yuyiguo:
@muhammadimranfarooqi 
Actually, We don't need to change the concurrent access number on k8s. So there is no needs to change it in gitlab. You may use the existing config for k8s.

muhammadimranfarooqi:
@yuyiguo 

The line corresponding to view_instances_k8s will be removed from deploy script. Because this file is available in gitlab. I am not changing concurrent access on k8s.

PR #930, opened 2020-07-07
Reduce silence tags to only have cmsweb; adjust maintenace alert

vkuznet:


vkuznet:
@muhammadimranfarooqi , I revisited again the silence part, it seems to me that it matches all given tags, and therefore I only left out tag=cmsweb to allow matching with cmsweb service alerts. Please merge and we'll try it out in next deployment upgrade.

PR #931, opened 2020-07-14
Add CRIC link upon request of Justas

vkuznet:


vkuznet:
@muhammadimranfarooqi  please include it in next cmsweb upgrade

PR #944, opened 2020-09-03
WIP: adds blacklisting capabilities to mkvomsmap script

Panos512:
I created two new files 'mkvomsmap_2' and `mkgridmap_2.conf` so that we can test the two implementations in parallel. 
When we verify that the new implementation works we can switch old script to the new implementation.

I have made a test private GitLab repo and hosted a file with one DN per line. The script fetches this file and excludes from the final map all the DNs that appear in the file.

We need to decide where we will put the blacklist file and then create an access token from a GitLab account that has access to that repo. The token is passed as a parameter in the script. We could probably add it in an environmental variable and distribute it through puppet to all machines.

I already tested in vocms0135 running:

`/data/srv/current/config/frontend/mkvomsmap_2 --key /data/certs/hostkey.pem --cert /data/certs/hostcert.pem -c /data/srv/current/config/frontend/mkgridmap_2.conf -o ./testout --git_token private_token`

### DIFF
--- a/frontend/mkvomsmap_2
+++ b/frontend/mkvomsmap_2
@@ -1,0 +1,168 @@
+#!/usr/bin/env python
+""" mkvomsmap -o OUTPUT --cert CERT --key KEY [--vo VONAME]
+                  [-c CONFIG-FILE] [--git-token GIT_TOKEN] [URI...]
+
+Queries VOMS for member data and writes out a gridmap file into OUTPUT
+as VONAME memberships (default: 'cms'). The CERT and KEY should normally
+be the host certificate and key, respectively. Queries all given URIs
+and combines all their results into one combined output list.
+
+Users who's DNs appear in the blacklist file are not included in the output.
+The blacklist file is maintained and hosted in GitLab and in order to access
+it a valid GIT_TOKEN must be provided.
+
+The output file is written in a manner which guarantees it will not be
+removed or truncated in case of a failure, however it may disappear for
+a very short time while the new file is moved in place.
+
+The URIs should be of the form:
+
+  vomss://voms2.cern.ch:8443/voms/cms?/cms/Role=production
+  vomss://voms2.cern.ch:8443/voms/cms?/cms
+
+If CONFIG-FILE is given, it's assumed to consist of lines in format
+'group URI VONAME', compatible with edg-mkgridmap, and a line in format
+'gitlab URL' to provide the url of the blacklist file.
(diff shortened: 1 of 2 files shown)


frontend/mkvomsmap_2
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python
+""" mkvomsmap -o OUTPUT --cert CERT --key KEY [--vo VONAME]
```
vkuznet, line 2: Panos, please update docstring to reflect changes about gitlab and blacklist. The documentation should clearly explain how it is done in a script and what it will do with a blacklist file. We also need to document content of blacklist.txt file here.

frontend/mkvomsmap_2
```diff
@@ -0,0 +1,161 @@
...
+        sys.exit(1)
+      uris.append(m.group(1))
+    if n:
+      gitlab_url = n.group(1) + '&access_token=' + opts.git_token
+
+
+try:
+  blacklisted_dns = urllib2.urlopen(gitlab_url).read().split('\n')
+except Exception:
+  blacklisted_dns = []
```
vkuznet, line 139: please add 
```
import traceback
traceback.print_exc()
```
to have clear message about what was wrong with gitlab_url. For instance, it can be a network problem or it can be file issue itself.

frontend/mkvomsmap_2
```diff
@@ -0,0 +1,161 @@
...
+for uri in uris:
+  users |= parse(request(uri))
+
+# Now write this out, first in a temporary file in the same directory as
+# the final output file, then move the output file into place. This avoids
+# wiping out the file in case the disk fills up or some other error occurs.
+fd, tmpname = mkstemp(dir = os.path.dirname(opts.out))
+outfile = os.fdopen(fd, "w")
+for user in users:
+  if user in blacklisted_dns:
```
vkuznet, line 153: I suggest to add comment here to stay explicitly that `user` is in fact is user DN.

vkuznet (changes requested):

Panos512:
Thanks a lot, @vkuznet. I fixed the issues. Also renamed the argument of the script from `--git_token` to `--git-token` to follow the usual convention.

vkuznet:
+1 from my side, please test and deploy with @muhammadimranfarooqi to testbed first and try to blacklist yourself to see if FE will behave properly and will not allow you to access cmsweb.

PR #952, opened 2020-09-08
updates mkvomsmap script to blacklist specific users

Panos512:
We completed the necessary testing with @muhammadimranfarooqi.

In order for the changes to be taken into account, a reload of the FE is necessary.
I suggest the following change in the cron so that the refresh of the file happens every 5 minutes and then immediately the FE Is restarted:

*/5 * * * * /data/srv/current/config/frontend/mkvomsmap --key /data/certs/hostkey.pem --cert /data/certs/hostcert.pem -c /data/srv/current/config/frontend/mkgridmap.conf -o /data/srv/state/frontend/etc/voms-gridmap.txt --vo cms --git-token {TOKEN}; sudo bashs -l -c "/data/srv/current/config/frontend/manage restart 'I did read documentation'"

Before merging that we should decide on the repository the config file will rest. After that, the configuration must change to point to that file.


Panos512:
I have been investigating using the `revoked-users.txt` file to blacklist people. My tests have failed and I don't find any references to that file in the CMSWeb code or here so that I can debug it. 

I propose we proceed with this MR:
1. Leaving the removal of DNs as part of the default functionality.
2. Not altering the cron but instead every time we want to blacklist we have to:
        a. Update the file on GitLab.
        b. SSH in the FE, manually run the cron and restart the FE.

What do you think? Should I proceed with that plan?

vkuznet:
Panos,
that was our original plan and I don't see why you can't proceed with that.
Please go ahead.

amaltaro:
Note that CMSWEB has many frontends and leaving it to a manual action might be error prone.
How about detecting - in the script executed in the cronjob (mkvomsmap?) - when the list of blacklisted users changed, and then automatically triggering an apache restart?

vkuznet:
Alan,
I doubt it is feasible in current infrastructure. There are only two ways to accomplish this:
- modify apache itself to monitor a certain file and restart
- modify our manage script to add yet another daemon which will monitor a file and restart apache

The problem with former is that we don't want to adjust apache codebase, I doubt anyone has this knowledge. The problem with second one is that there is no guarantee that update of a file will have consistently on all node due to possible intermittent network error when you fetch file from gitlab, e.g. git may fail on one of the FE node. Therefore a manual procedure is a most robust way to do it and operator or anyone with proper access right can perform the action.

Panos512:
OK this is now tested and it works fine.
A new GitLab repo has been created to host the blacklisting file: https://gitlab.cern.ch/cmsweb/cmsweb-blacklisting 
The code has been configured to fetch the file from there.

After this is merged there are three things that need to be done:
1. Remove me and Muhammad from the blacklisting file :) 
2. Ammend the cron job command with a flag including a GitLab API token that has access to the repo `--gitlab-token {GITLAB_TOKEN}`. We can use my token but I would suggest Muhammad issues once since he is the CMSWeb operator.
3. Add documentation to the new CMSWeb documentation site. I'll do that soon.

vkuznet:
Panos, we need to test this procedure on k8s too. Please do so and report.

amaltaro:
@Panos512 Hi Panos, it looks like this PR breaks the VM-based private deployments. Can you please check the validation gitlab ticket (#234) and work on a fix for such cases? Thanks

vkuznet:
@Panos512 , I think what should be done is to check if given token file exists, if not then it should not be applied. This way it will be backward compatible on private VMs.

Panos512:
Sorry about that @amaltaro! I thought this file is put there by puppet but I definitely shouldn't have assumed that :) 
I think #959 fixes it. If an `IOError` is thrown gitlab_url will not be amended correctly and it will just display a warning, proceeding with the file creation

PR #975, opened 2020-11-03
Disable Reqmon CherryPy thread tasks from pre-production VMs

### LINKED ISSUE #973: Disable reqmon cherrypy threads
Do not run CherryPy threads in deployment config

Reason: These will be running from the k8s deployment instead (and we can't run 2 at the same time).

https://github.com/dmwm/deployment/blob/master/reqmon/config.py#L93

khurtado:
Fixes #973


amaltaro:
Together with this change, we need to make sure `reqmon` gets deployed in k8s-testbed (which I think is already the standard procedure):
https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/services/reqmon.yaml

amaltaro:
LGTM!

PR #980, opened 2020-11-09
Disable pre-prod reqmgr2 CherryPy threads

goughes:
We are migrating reqmgr2 to the k8s testbed. We cannot run VM and k8s reqmgr2 threads at the same time so this PR will prevent the VM instances of the reqmgr2 cherrypy threads from running.

### DIFF
--- a/reqmgr2/config.py
+++ b/reqmgr2/config.py
@@ -123,7 +123,9 @@ ui_main.application = ui.index
 
 extentions = config.section_("extensions")
 # Production instance of wmdatamining, must be a production back-end
-if HOST.startswith("vocms0766") or HOST.startswith("vocms0731") or HOST.startswith("vocms0117"):
+# Disable CherryPy thread tasks from testbed (vocms0731), as this will be started
+# from k8s and only one cherrypy instance can be run at a time to avoid issues with CouchDB
+if HOST.startswith("vocms0766") or HOST.startswith("vocms0117"):
 #     wmdatamining = extentions.section_("wmdatamining")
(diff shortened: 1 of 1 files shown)


muhammadimranfarooqi:
@goughes 

Could you please resolve conflict. Thanks

goughes:
Sure, let me rebase real quick...

goughes:
Ok, looks good now from my side.

amaltaro:
@goughes Erik, since this disables the reqmgr2 cherrypy threads, it means we should deploy reqmgr2-tasks in k8s.
However, what we should do in the beginning of this week, was to just enable reqmgr2 in k8s-testbed, which means, no cherrypy threads are needed (what we need is to make sure `reqmgr2-tasks` doesn't get deployed in k8s).

This is what we defined in that timeline table, here:
https://gitlab.cern.ch/cms-http-group/doc/-/issues/236

PS.: I know it's confusing ;)

goughes:
Yep, you're right. I overlooked the reqmgr2-tasks entry in the table and was thinking reqmgr2 itself included these threads. Guess we can park this PR until later this month. Thanks Alan!

amaltaro:
This is to be merged between Nov 20th and 25th.

amaltaro:
Just noticed it has a conflict! Will have to fix it, I guess in a new PR.

amaltaro:
Conflicts have been resolved in a new PR: https://github.com/dmwm/deployment/pull/992
Feel free to close this PR.

PR #983, opened 2020-11-17
drop phys01/2 deployment from DBS instances.

yuyiguo:


muhammadimranfarooqi:
@yuyiguo 

Could you please also make similar changes for `services_config` repository both for `preprod` and `prod` branches. Thanks

yuyiguo:
I will.
Let's finish the VM first and make sure that works properly.

PR #987, opened 2020-11-23
Disable production reqmon CherryPy threads

goughes:
This PR will disable reqmon production cherrypy threads in the VM infrastructure. These threads will migrate to running via a reqmon-tasks pod in production k8s on Nov 24.


amaltaro:
+1 (keep in mind we have to *first* disable it  in the VMs, only then enable in k8s)

PR #989, opened 2020-11-23
Disable testbed VM instance of workqueue

goughes:
This PR will disable workqueue in the testbed VM infrastructure. Workqueue will migrate to running in testbed k8s on Nov 24.


amaltaro:
+1 (keep in mind we have to *first* disable it  in the VM testbed, only then enable it in k8s-testbed)

PR #994, opened 2020-11-24
Do not remove MongoDB databases at sysboot

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/10053

I fail to see a good reason to have that "rm -rf" in place... which was added by Lassi 10 years ago...

@vkuznet Valentin, please let me know if you have any input on this 1 line change. Otherwise, I'm pretty sure that's the line that screwed with MSOutput in production(!)

### DIFF
--- a/mongodb/manage
+++ b/mongodb/manage
@@ -45,7 +45,7 @@ sysboot()
   PID=$(cat $STATEDIR/db/mongod.lock 2>/dev/null)
   if [ X"$PID" = X ] || [ $(ps -o pid= -p $PID | wc -l) = 0 ]; then
     # must clean-up improper shutdown
-    rm -rf $STATEDIR/db/*
+    rm -f $STATEDIR/db/mongod.lock
     rm -f $PWD/stagingarea/*-schema-stamp
     start
   fi
@@ -61,7 +61,7 @@ start()
       cmd=""
   fi
   $cmd $MONGO_ROOT/bin/mongod --dbpath=$STATEDIR/db --storageEngine mmapv1 \
-      --port 8230 --quiet --nojournal --nohttpinterface --bind_ip 127.0.0.1 \
+      --port 8230 --quiet --nohttpinterface --bind_ip 127.0.0.1 \
       </dev/null 2>&1 | rotatelogs $LOGDIR/mongodb-%Y%m%d-`hostname -s`.log 86400 >/dev/null 2>&1 &
 
   n=0 started=false


vkuznet:
Alan, this line is important. Here is use-case:
- if mongodb is not properly shutdown, e.g. when node is rebooted or when it is killed, it will leave the lock file.
- the lock file is a confirmation that there is either existing MongoDB instance running or it was an instance which was not properly shutdown
- therefore, the DB may be in corrupted state and it is easier to start with a clean state, that's why there is rm.

But that was done as you said 10 years ago, since then MongoDB team added a journal for DB, like ext2 vs ext3 on Linux system. The former was fast but left OS unclean, while later had a filesystem journal which can recover all transactions against file-system. As such, if MongoDB is run with journal it can recover, otherwise it may have corrupted files. Therefore, to make proper change you can't just remove this line, instead, you can remove this line only if you add journal to MongoDB process. It is normal write-ahead feature of DB, please read about it here: https://docs.mongodb.com/manual/core/journaling/

For DAS we need a fast DB and since we don't care about persistency we don't really need a journal. For WMCore use case you may need journal since you may not need the fastest DB. The two use-cases are opposite and we should properly take care of this in manage script. 

Here is DAS mongodb.conf file:
```
systemLog:
   destination: file
   path: "/data/mongodb/logs/mongod.log"
   logAppend: true
   logRotate: rename
storage:
   dbPath: "/data/mongodb/wtdb"
   journal:
      enabled: false
processManagement:
   fork: true
net:
   bindIp: 0.0.0.0
   port: 8230
```

And, in your case I think you'll need to enable journal. I suggest to do the following:
- read mongodb conf file
- if journal is set to true then we can do not delete db files
- while if journal is false we delete db files

Feel free to make appropriate changes and I can review them once they will be ready.

amaltaro:
Valentin, my understanding from your reply is that deleting ALL the mongodb databases is not the important piece, but just the lock file `mongod.lock`. How about we delete only that file at sysboot then and leave all the rest of the configuration/mode as is?

About `journal` or not, the process is configured to start with `--nojournal` option, see:
https://github.com/dmwm/deployment/blob/master/mongodb/manage#L64

did you ever check the performance difference for DAS with and without journal? If it's only DASWeb that uses the mongodb as a caching, and most of the users are actually using the DAS client, I wonder if we shouldn't just try running mongo with journaling enabled?

PS.: from a quick look at the documentation you pointed out, it seems `--nojournal` is not supported in newer MongoDB versions (and we are likely going to upgrade it soon, since we are using an EOL'ed version at the moment). So it's natural to me that we start enabling journaling as well.

amaltaro:
According to that documentation, there seems to be some special notes for mongodb in journaling mode (writes between operations, checkpoints, etc). So I'd suggest to enable that in the next CMSWEB release.

About the databases removal, I updated this patch to only delete the lock file. Deleting ALL the databases is totally pointless.

vkuznet:
Alan,
yes I did check performance with and without journal, and enabling journal slow down the data injection rate. But this was done with original DB backend used by MongoDB based on memory mapped file. This is also the case of `--nojournal` option which was used in earlier versions of MongoDB. Now MongoDB switched to another backend, called WiredTiger and I use in DAS. For that, the journal option is specified in mongodb configuration as I provided above.

DAS uses case is certainly different from WMCore, since DAS needs fastest IO to inject lots of records from ongoing query. The latency is visible on cmsweb. In WMCore you don't need 20kHz (the rate I measured with nojournal) of injection rate.

Now, about lock and DB file. I think you still mistaken about it. If you delete lock it means there were a MongoDB process with left DBs in some state. If you don't use journal it implies that the state of DB is unknown, i.e. it can be in a good or bad state. The journal guarantee that DB can be recovered from a bad state. Therefore removal of lock and leaving DB can only be safely done if you *use* journal. And, as I wrote before it is not the case of DAS. The good news is that DAS no longer runs on VM cluster, and no longer need manage script since in k8s it is much simpler procedure. Therefore, you can do to mongo manage script whatever you want and it will not affect DAS. But as I wrote be aware of your actions.

amaltaro:
I see. I'm going to update this PR then and run some quick tests without the `--nojournal` option.
I failed to find the mongod.conf in the cmsweb VM backend, do you know where it's supposed to be?

vkuznet:
SO far I keep it here
https://github.com/dmwm/CMSKubernetes/blob/master/docker/das-mongo/mongodb.conf
But probably should move it to gitlab as any other configuration. Please note that this configuration file is for DAS MongoDB use case.

vkuznet:
ok, and now it is available on gitlab for k8s, https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/prod/das-mongo/mongodb.conf

amaltaro:
I see you also pull a newer version of mongodb 4.2.8. I guess your services no longer rely on mongodb specs in cmsdist, right?

vkuznet:
yes, I'm testing now 4.4.2, and with k8s setup I no longer use RPMs in any services I maintain, and it also means I no longer need anything from cmsdist and/or development.

vkuznet:
+1

PR #997, opened 2020-11-26
Fix certificate key owner for workqueue/reqmgr2ms

amaltaro:
Complement to: https://github.com/dmwm/deployment/pull/996

We need to update the ownership as well, since it's owned by cmsweb and the service account isn't able to load the key file.


amaltaro:
none of the options we tried worked well. Trying to change things from the manage script does not work either. We decided to update the permissions and ownership during the production deployment, until those services have migrated to k8s.
Closing, thanks Imran for helping with the tests.

PR #998, opened 2020-12-02
New redirect rules for dbs2go

vkuznet:
New FE rules to allow testing DBS GoLang implementation


vkuznet:
@muhammadimranfarooqi , could you please merge this PR and create new FE image with it. I need to get new redirect rules for FE at least in testbed to start testing GoLang implementation of dbs.

muhammadimranfarooqi:
@vkuznet 

May be an entry here would also be required for these rules.

https://github.com/dmwm/deployment/blob/master/frontend/backends-k8s-preprod.txt

vkuznet:
Good point, I updated this PR.

PR #999, opened 2020-12-02
Change order of dbs redirects

vkuznet:
@muhammadimranfarooqi , turns out that original rules didn't work since /dbs was picked first instead of /dbs2go. I changed the order of rules. Please merge and redeploy again FE's on k8s testbed


muhammadimranfarooqi:
@vkuznet 

I have created new image and deployed again. Still does not work

muhammadimranfarooqi:
I think FE rule is correct but inside service there is issue.

```
[root@dbs2go-cd8c88ccc-jqgbh data]# curl localhost:8258/dbs2go
404 page not found
```
Same response comes from URL https://cmsweb-testbed.cern.ch/dbs2go/

vkuznet:
@muhammadimranfarooqi , I fixed the prefix in dbs2go and now it works.

PR #1000, opened 2020-12-03
Changed email which gets reports

jfernan2:
to cms-dqm-coreTeam@cern.ch


jfernan2:
Hi @muhammadimranfarooqi @amaltaro 
We would need to have this change in deployment at your earliest convenience.
Thanks

muhammadimranfarooqi:
@jfernan2 

I have applied these changes directly in the VMs running DQMGUI. Otherwise, with this PR, these were included in the next month upgrade.

jfernan2:
Thank you @muhammadimranfarooqi 
Indeed much faster this way.

PR #1002, opened 2020-12-07
Disable production VM reqmgr2 CherryPy threads

goughes:
This will disable the reqmgr2 CherryPy threads in the production VM infrastructure. These reqmgr2 threads should be disabled before the reqmgr2-tasks pod is enabled in k8s production on Dec. 8


amaltaro:
This looks good to me, thanks!

@muhammadimranfarooqi note that we *must* first update this configuration on vocms0766 and restart reqmgr2; only then we can create the reqmgr2-tasks pod in Kubernetes production.

PR #1003, opened 2020-12-07
Disable production VM instance of workqueue

goughes:
This will disable the prod VM instance of workqueue as part of the k8s migration. The VM instance should be disabled before starting the k8s workqueue pod. Workqueue migration is scheduled for Dec. 10.


amaltaro:
This looks good to me as well, thanks!

@muhammadimranfarooqi note that we **must** first update this configuration on vocms0740 and restart workqueue; only then we can create the workqueue pod in Kubernetes production. Please ping me on gtalk to check the service on the VM-based node before enabling it in k8s.

amaltaro:
This cannot be merged yet!

We need to make changes to WMCore before we can migrate global workqueue to kubernetes.

PR #1006, opened 2020-12-10
Disable testbed VM instances of ms-transferor, ms-monitor and ms-rulecleaner

goughes:
This PR will disable the ms-transferor/monitor/rulecleaner microservices on the VM testbed. This should be merged before deploying the transferor, monitor and rulecleaner pods as part of their migration to the testbed k8s infrastructure.

Original date for this migration was Dec 9, but we did not get things ready by then. Will update this PR with a new date for the migration after a discussion with @amaltaro

### DIFF
--- a/reqmgr2ms/manage
+++ b/reqmgr2ms/manage
@@ -34,10 +34,22 @@ ROOT=$(cd $(dirname $0)/../.. && pwd)
 CFGDIR=$(dirname $0)
 LOGDIR=$TOP/logs/$ME
 STATEDIR=$TOP/state/$ME
-CFGFILETR=$CFGDIR/config-transferor.py
-CFGFILEMON=$CFGDIR/config-monitor.py
-CFGFILEOUT=$CFGDIR/config-output.py
-CFGFILERCLEAN=$CFGDIR/config-ruleCleaner.py
+
+# Disable specific microservices in VM testbed for k8s migration
+# This will cause config file checks for transferor, monitor and
+# ruleCleaner to return false and services will not start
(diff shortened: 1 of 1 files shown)


amaltaro:
I think this one cannot be merged yet as well.

@goughes unless I missed something, I think we need to make a manage script for each of the microservices to be deployed in kubernetes (such that they act only on the microservice they are supposed to).

goughes:
@amaltaro I believe that all the microservice config files from the deployment repo will be removed via this: https://github.com/dmwm/CMSKubernetes/blob/df2be73b1a5775e62395328a2402d5da3f8483f5/docker/reqmgr2ms/run.sh#L50

Leaving only the single microservice config file from /etc/secrets (what is in services_config). The logic in the manage script will only start a microservice if the config file exists, so we should only see one microservice start based on which config file is present. @muhammadimranfarooqi does that sound right?

However, this will need to be updated to include the ruleCleaner config file: https://github.com/dmwm/CMSKubernetes/blob/df2be73b1a5775e62395328a2402d5da3f8483f5/docker/reqmgr2ms/run.sh#L53

amaltaro:
Ohh, that's a good analysis and I think you're right. However, to avoid BAD surprises, I'd like to have a confirmation of this behaviour from Imran.

Erik, can you get us a PR for that run.sh script please?

muhammadimranfarooqi:
@amaltaro @goughes 

Erik's analysis is right.

todor-ivanov:
Hi @goughes this looks good too. Thanks for tracking the needed logic here.

amaltaro:
Thanks everyone. It looks good to me too!

PR #1010, opened 2020-12-15
Add parameters needed for including the archival logic inside MSRuleCleaner

todor-ivanov:
Those are the parameters needed for moving the archival logic from the cherrypy treads into the MSRulecleaner service.


amaltaro:
LGTM!

PR #1013, opened 2020-12-17
Added Rucio Consistency Monitor

ivmfnal:


muhammadimranfarooqi:
@ivmfnal 

You cannot use `9201` port as it is already being used by `exitcodes` service. Can you use `9202` instead?

ivmfnal:
I thought I was using port 8400 as Valentin suggested.
Where did I put 9201 ?
________________________________

ivmfnal:
Ah yes, I see it now. It's in app_rucioconmon_ssl.conf.
Sorry I did not notice that. It was copied from errorcodes file. Can you replace it with 8400 please ?
Thanks !

muhammadimranfarooqi:
Please replace port with the new commit. 

Also add entry `^/auth/complete/rucioconmon(?:/|$) <servicename>.<namespace>.svc.cluster.local` in this file

https://github.com/dmwm/deployment/blob/master/frontend/backends-k8s-preprod.txt

and `^/auth/complete/rucioconmon(?:/|$) cmsweb-srv.cern.ch` in  this file
https://github.com/dmwm/deployment/blob/master/frontend/backends-k8s-preprod.txt

ivmfnal:
What exactly those files are, anyway ? Are they documented anywhere ?

ivmfnal:
What is namespace here ?

muhammadimranfarooqi:
These files define the redireciton rules for apache frontend. The `backends-k8s-preprod.txt` is for the testbed k8s 
cluster and `backends-k8s-prod.txt` is for the production k8s cluster.

muhammadimranfarooqi:
The `namespace` in the kubernetes cluster defines a scope in which you run your service. If you did not use namespace in your service, then you can use `default`. Infact, better to define and use namepace something like `rucio`

muhammadimranfarooqi:
Infact you used `http` in your manifest file. 

https://github.com/dmwm/CMSKubernetes/pull/424/files#diff-37e667ab4d24af91c3c3ece19bdab05dd3ab02b90d00102d7d4d92c7471a47fdR7

ivmfnal:
So the namespace is for me (us) to choose arbitrarily or it is assigned by you guys ?

Igor
________________________________

muhammadimranfarooqi:
yes, you can use any name with respect to its scope. However, you used `http` in the manifest file. If you change there then the same will need to be put here as well.

ivmfnal:
Ok, I am going to change the namespace

ivmfnal:
If I want to use "rucio-cm" as the namespace, do I need to do anything to create it ?

muhammadimranfarooqi:
The namespace and service name should be consistent with the k8s manifest file. 

But here, these are different. 

https://github.com/dmwm/CMSKubernetes/pull/424/files#diff-37e667ab4d24af91c3c3ece19bdab05dd3ab02b90d00102d7d4d92c7471a47fdR51-R52

ivmfnal:
The corrected namespace/name are in a separate PR: https://github.com/dmwm/CMSKubernetes/pull/426

vkuznet:
@muhammadimranfarooqi now I think they are consistent, in this PR it is monitor.ruciocm.xxx and in dmwm/CMSKubernetes#426 the app is monitoring and namespace is ruciocm

muhammadimranfarooqi:
@ivmfnal 

I have deployed new frontend image with this PR changes in `cmsweb-test4` cluster.  Your application should be accessible now in this cluster. Please test. Thanks

muhammadimranfarooqi:
@ivmfnal 

If I try to access cmsweb-test4.cern.ch/rucioconmon

```
< HTTP/1.1 502 Proxy Error
< Date: Fri, 18 Dec 2020 11:21:18 GMT
< Server: Apache
< Set-Cookie: cms-auth=ff5ee0d14374261d793bdb2f01fb9e9c822f3bead4d73f23342d5dfd2e39f2ec49f4dfcdf9cf8060;path=/;secure;httponly;expires=Thu, 01-Jan-1970 00:00:01 GMT
< Content-Length: 454
< Content-Type: text/html; charset=iso-8859-1
<
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>502 Proxy Error</title>
</head><body>
<h1>Proxy Error</h1>
<p>The proxy server received an invalid
response from an upstream server.<br />
The proxy server could not handle the request <em><a href="/auth/complete/rucioconmon">GET&nbsp;/auth/complete/rucioconmon</a></em>.<p>
Reason: <strong>DNS lookup failure for: monitor.ruciocm.svc.cluster.local</strong></p></p>
</body></html>
* Connection #0 to host cmsweb-test4.cern.ch left intact
```

However, if I look at the namespace in test4 cluster it is `rucio-cm` while you used in the path `ruciocm`. You may adjust namespace name in the cluster to `ruciocm` and then redeploiy you app. Then it should work.

PR #1015, opened 2020-12-22
Add new FE rules for imagebot

vkuznet:
This is experimental FE rules for imagebot.

@muhammadimranfarooqi the imagebot will use its own token based authentication, therefore I don't put any redirection to FE auth module since we can't use x509 for that. The imagebot rules are based on https and token. Please put these rules on testbed that I can check whole workflow. We can avoid building FE image and/or RPMs with them and instead take existing config on testbed FE node, put it somewhere to add these rules and then apply it via secrets to FE pods.


vkuznet:
please discard this PR, since imagebot only required in k8s it is better to adjust frontend install.sh script in CMSKubernetes/docker/frontend area.

PR #1019, opened 2021-01-12
Adjusting deploy scripts for new DBS nodes

muhammadimranfarooqi:


muhammadimranfarooqi:
@amaltaro 

As discussed today, could you please review this PR. Thanks

crabserver/deploy
```diff
@@ -38,7 +38,7 @@ deploy_crabserver_sw()
 
 deploy_crabserver_post()
 {
-  case $host in vocms013[89] | vocms073[89] | vocms0143 | vocms074[0123456] | vocms0307 | vocms0318 ) disable ;; * ) enable ;; esac
+  case $host in vocms013[89] | vocms073[89] | vocms0143 | vocms074[0123456] | vocms076[12345] | vocms0307 | vocms0318 ) disable ;; * ) enable ;; esac
```
amaltaro, line 41: crabserver is a general purpose service, so it could likely run from these nodes too.

phedex/deploy
```diff
@@ -35,7 +35,7 @@ deploy_phedex_sw()
 deploy_phedex_post()
 {
   case $host in
-    vocms013[89] | vocms073[89] | vocms0143 |  vocms074[0123456] | vocms0307 | vocms0318 )
+    vocms013[89] | vocms073[89] | vocms0143 |  vocms074[0123456] | vocms0307 | vocms0318 |  vocms076[12345] )
```
amaltaro, line 38: even though PhEDEx datasvc will soon be shutdown, it is a general purpose service, so it could likely run from these nodes too.

reqmgr2/deploy
```diff
@@ -44,7 +44,7 @@ deploy_reqmgr2_post()
 {
   # in practice, general purpose backends, couchapps on vocms0132 (preprod) and vocms0742 (prod)
   case $host in
-    vocms013[89] | vocms073[89] | vocms074[013456] )
+    vocms013[89] | vocms073[89] | vocms074[013456] |  vocms076[12345]  )
```
amaltaro, line 47: reqmgr2 is a general purpose service, so it could likely run from these nodes too.

reqmon/deploy
```diff
@@ -44,7 +44,7 @@ deploy_reqmon_post()
 {
   # in practice, general purpose backends, couchapps on vocms0731 (preprod) and vocms0743 (prod)
   case $host in
-    vocms013[89] | vocms073[89] | vocms074[012456] )
+    vocms013[89] | vocms073[89] | vocms074[012456] |  vocms076[12345]  )
```
amaltaro, line 47: reqmon is a general purpose service, so it could likely run from these nodes too.

t0_reqmon/deploy
```diff
@@ -42,7 +42,7 @@ deploy_t0_reqmon_post()
 {
   # in practice, general purpose backends, couchapps on vocms0132 (preprod) and vocms0744 (prod)
   case $host in
-    vocms013[89] | vocms073[89] | vocms074[012356] )
+    vocms013[89] | vocms073[89] | vocms074[012356] |  vocms076[12345] )
```
amaltaro, line 45: t0_reqmon is a general purpose service, so it could likely run from these nodes too.

amaltaro (commented):
Imran, I made some comments along the code. Given that most - or all - of those services are already running on k8s, it shouldn't be a big deal whether we enable or not them in the new nodes.

However, if we ever use them from the VM nodes again, you might need to consider the general frontend rules (backend-prod map) for the general purpose services/backends.

muhammadimranfarooqi:
@amaltaro 

Although, new nodes are general purpose backends, but they are supposed to be used by DBS production clients. So, only DBS services will be running in these nodes. We'll have new FE clusters using alias `cmsweb-prod.cern.ch` that will point to new DBS nodes.

amaltaro:
I see, then it should be all fine! Thanks Imran.

PR #1020, opened 2021-01-15
Disable ms-output in prod and testbed VMs

goughes:
ms-output was already stopped manually in the VM testbed when ms-output was migrated to preprod k8s. This would ensure it does not start.

cc @amaltaro


amaltaro:
It looks good to me Erik. However, I think we can go even a step further and have it already disabled for the production node as well. What do you think?

amaltaro:
Thanks Erik, it's good to go from my side!

PR #1022, opened 2021-01-20
Commenting kerberos lines

jfernan2:
Kerberos lines are giving problems in lxplus installations at present January 2021
when trying to load any DQM GUI instance, commenting them works perfectly fine since users alrady have a kerberos token by default

### DIFF
--- a/dqmgui/manage
+++ b/dqmgui/manage
@@ -69,13 +69,16 @@ export RFIO_USE_CASTOR_V2=YES
 export STAGE_SVCCLASS=archive
 
 # kerberos
-export keytab=/data/srv/current/auth/wmcore-auth/keytab
-principal=`klist -k "$keytab" | tail -1 | awk '{print $2}'`
-kinit $principal -k -t "$keytab" 2>&1 1>& /dev/null
-if [ $? == 1 ]; then
-    echo "Unable to perform kinit"
-    exit 1
-fi
+# following lines are giving problems in lxplus installations at present January 2021
+# when trying to load any DQM GUI instance
+# commenting them works perfectly fine since users alrady have a kerberos token by default
+#export keytab=/data/srv/current/auth/wmcore-auth/keytab
+#principal=`klist -k "$keytab" | tail -1 | awk '{print $2}'`
+#kinit $principal -k -t "$keytab" 2>&1 1>& /dev/null
+#if [ $? == 1 ]; then
+#    echo "Unable to perform kinit"
+#    exit 1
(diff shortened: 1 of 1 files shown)


jfernan2:
Please merge at your earliest convenience. Thanks

muhammadimranfarooqi:
@jfernan2 

How does it work? Its `cmsweb` user that requires `kinit` authentication to get access to EOS. Which error are you getting?

jfernan2:
Hi @muhammadimranfarooqi 
The users (I reproduce the error too) get the following error:

klist: Key table file '/data/srv/current/auth/wmcore-auth/keytab' not found while starting keytab scan
Unable to perform kinit

Indeed the keytab file in that address does not exist in lxplus, I am not sure if it has changed somehow in lxplus but, removing tjhose lines users are able to load a test DQM GUI, so my feeling is this is not needed any longer
Thanks

muhammadimranfarooqi:
@jfernan2 

Ok let me check and I'll get back to you

muhammadimranfarooqi:
@jfernan2 

It does exist in all VMs for DQM. 

```
[cmsweb@vocms0731 ~]$ ls /data/srv/current/auth/wmcore-auth/
header-auth-key  keytab

[cmsweb@vocms0739 ~]$ ls /data/srv/current/auth/wmcore-auth/
header-auth-key  keytab

[cmsweb@vocms0738 ~]$ ls /data/srv/current/auth/wmcore-auth/
header-auth-key  keytab
```
These were added by @ahmad3213 during last month deployment and these worked fine.

jfernan2:
OK, I understand, so this is the reason why is failing now in lxplus for users trying to install a test GUI.
Since users always get the master version of the file, I am not sure how to proceed: I guess the easiest solution is to put a comment in the file so that users modify the file for their local instalations

muhammadimranfarooqi:
@jfernan2 

The other way would be for users to generate keytab file for themselves and put it in the desired location. The procedure is available at this doc https://cms-http-group.docs.cern.ch/vm_cluster/routine_maintenance/#generate-keytab-file

jfernan2:
Thanks @muhammadimranfarooqi 
I have quoted those instructions in the installation twiki:
https://twiki.cern.ch/twiki/bin/view/CMS/DQMGuiForUsers#How_to_set_up_your_own_DQM_GUI_f
If you can merge this PR with the inocent comment and updated warning message to make everybody aware, it would be great.
Thanks

PR #1032, opened 2021-02-17
Disable caching when retrieving bootstrap.sh file

andrius-k:
Proxy servers might cache the old version of the file. This change disables caching.

### DIFF
--- a/Deploy
+++ b/Deploy
@@ -246,7 +246,7 @@ inrepo()
      rm -fr $swarea
      mkdir -p $swarea
      cd $swarea
-     curl -sO http://$rpmhost/cmssw/bootstrap.sh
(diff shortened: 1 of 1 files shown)


muhammadimranfarooqi:
@vkuznet 

Could you please review. Thanks

Deploy
```diff
@@ -246,7 +246,7 @@ inrepo()
      rm -fr $swarea
      mkdir -p $swarea
      cd $swarea
-     curl -sO http://$rpmhost/cmssw/bootstrap.sh
+     curl -sO -H 'Cache-Control: no-cache' http://$rpmhost/cmssw/bootstrap.sh
```
vkuznet, line 249: well, if you want to be completely sure, just add block ti delete local `bootstap.sh` before calling the curl, e.g.
```
if [ -f $PWD/bootstrap.sh ]; then
   rm $PWD/bootstrap.sh
fi 
```

PR #1038, opened 2021-03-23
Remove phedex_url from T0/WMAgent manage script

amaltaro:
Part of the PhEDEx deprecation, no longer read it from the configuration file and no longer passes it through when deployment a (T0)WMAgent service.
Required by: https://github.com/dmwm/WMCore/pull/10366

With this PR, we no longer need to define: PHEDEX_URL in the secrets file (actually, we should delete it from there!)


amaltaro:
@muhammadimranfarooqi can you please merge this one? Thanks

PR #1040, opened 2021-03-24
Update ReqMgr2MS MSManager path

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/10375


amaltaro:
Tested and it indeed resolves the problems we have seen in today's HG2104b upgrade.

PR #1072, opened 2021-07-28
uses CERN CA bundle to verify robot cert

Panos512:


vkuznet:
@Panos512 , sorry but I would like to request further changes. Usage of hardcoded paths should be avoided. Please change code to pass both robot and rootCA certificate files as optional parameters to the script. This will make deployment much more flexible.

Panos512:
Thank Valentin, I'll do that (I now have to make them optional).
Do we store the cron commands somewhere, so that I can change the cron to pass the files?

PR #1074, opened 2021-07-29
New rewrite rules for MSUnmerged and backends map

amaltaro:
WORK IN PROGRESS

PR #1075, opened 2021-07-29
Fix rewrite rules for MSUnmerged, supporting rses query string

amaltaro:
Complement to: https://github.com/dmwm/deployment/pull/1074

Queries like (tested with my own VM url):
https://cmsweb-testbed.cern.ch/ms-transferor/data/info?detail=true&rses=t2-t3-us
https://cmsweb-testbed.cern.ch/ms-transferor/data/status?detail=true&rses=t2-t3-us

and the same for `rses=t1` and `rses=t2-t3` seem to be redirecting the http call to the correct backend now.


amaltaro:
@muhammadimranfarooqi can you please review this. If it looks good for you, please merge and push it to testbed such that we can run some final tests.

PR #1088, opened 2021-08-26
Initial manage/config changes for CouchDB 3.1.2

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/10829

Let me try to document all the changes and why they have been made:
* `couchdb/default.ini`, `wmagentpy3/default.ini` and `tier0/default.ini` configuration file is required to start up CouchDB. This is a blind copy of what is provided by CouchDB itself.
* `local.ini` configuration file contains our specific CouchDB configuration, overriding whatever is in the `default.ini`
  * `chttpd/authentication_handlers` define the Proxy Authentication mechanism
  * `couchdb/single_node` defined to true to get the admin users automatically created during deployment, in addition to some system-like databases
* `manage` script required a few changes, like:
  * database/management calls need to be made with user/pass, even if it's all localhost!
  * `couchdb -p`, `couchdb -a`, `couchdb -o`, `couchdb -e` and `couchdb -b` no longer exist, removed!
  * `couchdb -A` has been replaced by `couchdb -couch_ini`
  * adapted the `start` and `stop` code
  * CouchDB `status` now uses the `_up` unathenticated endpoint
  * pushing couchapps follow exactly the same logic, only difference is that it uses for loop now
* `couchdb/monitoring.ini` needs to use user/pass (set up during couchdb deployment)
*  `wmagentpy3/deploy` position CouchDB config files under the `couchdb` configuration directory
* `wmagentpy3/local.ini` defines 2 accounts, the admin one and unittest users. The unittest user will be disabled when CouchDB gets deployed in production. These accounts are automatically created during start up.
* `wmagentpy3/manage` is a mess and it makes our CI jenkins hard to deal with. 
* for central CouchDB deployment and `manage` script operations, read credentials from `couch_creds` file.
* define `max_replication_retry_count` to infinity for WMAgent/Tier0 backend
* define replication `connection_timeout` to 300 secs (I've seen timeouts in testbed replications!)

In addition to that, it also includes:
* changes to `reqmgr2ms/manage` such that MSOutput and MSUnmerged do not get started in our VMs (MongoDB setup is broken).

TODO: Imran, we need to define a file for the Couch credentials under /data/auth/couchdb. We can discuss this over Mattermost.

### DIFF
--- a/couchdb/manage
+++ b/couchdb/manage
@@ -37,14 +37,22 @@ CFGDIR=$(cd $(dirname $0) && pwd)
 LOGDIR=$TOP/logs/$ME
 STATEDIR=$TOP/state/$ME
 KEYFILE=$ROOT/auth/$ME/hmackey.ini
+COUCH_CREDS=$ROOT/auth/$ME/couch_creds
+EXCEPTIONS="wmstats"
+
 COLOR_OK="\\033[0;32m"
 COLOR_WARN="\\033[0;31m"
 COLOR_NORMAL="\\033[0;39m"
 PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/kerberos/bin
-export STAGE_HOST=${STAGE_HOST:-castorcms}
-export STAGE_SVCCLASS=archive
 
-EXCEPTIONS="wmstats"
+# Does the credential file exist and is it readable?
+if [ ! -r $COUCH_CREDS ]; then
+   echo "Variable COUCH_CREDS needs to be defined"
+   exit 1
+fi
+COUCH_USER=`cat $COUCH_CREDS | grep COUCH_USER | sed s/COUCH_USER=//`
+COUCH_PASS=`cat $COUCH_CREDS | grep COUCH_PASS | sed s/COUCH_PASS=//`
+COUCH_CREDS_URL="http://$COUCH_USER:$COUCH_PASS@localhost:5984"
 
 . $ROOT/apps/$ME/etc/profile.d/init.sh
 
@@ -195,7 +203,8 @@ files_needed_from_yui=('build/animation/animation-min.js'
 # Start service conditionally on crond restart.
 sysboot()
 {
-  if couchdb -p $STATEDIR/couchdb.pid -s; then :; else
+  num_procs=$(get_num_couch_process)
+  if [ -z "$num_procs" ]; then
     start
   fi
 }
@@ -204,10 +213,11 @@ sysboot()
 start()
 {
   cd $STATEDIR
-  couchdb -A $CFGDIR/ -a $KEYFILE -b \
-    -o /dev/null \
-    -e $LOGDIR/$(date +%Y%m%d-%H%M%S).stderr \
-    -p $STATEDIR/couchdb.pid </dev/null >/dev/null 2>&1
+  echo -n "Which couchdb: "
+  which couchdb
+  echo "  With configuration directory: $CFGDIR"
+  echo "  With logdir: $LOGDIR"
+  nohup couchdb -couch_ini $CFGDIR >> ${LOGDIR}/couch.log 2>&1 &
 
   push_apps
   replications push
@@ -216,18 +226,28 @@ start()
 # Stop the service.
 stop()
 {
-  couchdb -d -p $STATEDIR/couchdb.pid
+  echo "Stopping CouchDB service..."
+  # FIXME: grepping for couchdb31 is fragile!!!
+  for couch_pid in $(ps aux | grep couchdb31 | grep -v grep | awk '{print $2}'); do
+    echo "  killing CouchDB process... ${pid}"
+    kill -9 $couch_pid || true
+  done
+}
+
+# check the number of beam/couchdb processes
+get_num_couch_process()
+{
+  local num_proc=`ps aux | grep beam | grep -v grep`
+  echo $num_proc
 }
 
 # Check if the server is running.
 status()
 {
-  couchdb -p $STATEDIR/couchdb.pid -s
-
-  curl -s localhost:5984 | grep -q '"version":"1.[0-9.]*"' ||
+  curl -s localhost:5984/_up | grep -q '"status":"ok"' ||
     { echo "CouchDB is not correctly responding to requests"; return; }
 
-  local TASKS=$(curl -s localhost:5984/_active_tasks)
+  local TASKS=$(curl -s $COUCH_CREDS_URL/_active_tasks)
   [ "$TASKS" = '[]' ] && TASKS="No active tasks (e.g. compactions)"
   echo $TASKS
 
@@ -242,7 +262,7 @@ clean_views()
   [ -n "$database" ] ||
     { echo "You must specify the database you wish to clean the views "; exit 1; }
 
-  curl -s -H "Content-Type: application/json" -X POST http://localhost:5984/$database/_view_cleanup | \
+  curl -s -H "Content-Type: application/json" -X POST $COUCH_CREDS_URL/$database/_view_cleanup | \
        grep -q '{"ok":true}' ||
     { echo "An error occured while cleaning the views. Please look in the CouchDB logs."; exit 3; }
 }
@@ -259,8 +279,7 @@ push_apps()
 
   n=0 started=false
   while [ $n -le 100 ]; do
-    couchdb -p $STATEDIR/couchdb.pid -s &> /dev/null &&
-      curl -s localhost:5984 | grep -q '"version":"1.[0-9.]*"' &&
+    curl -s localhost:5984/_up | grep -q '"status":"ok"' &&
       started=true && break
     echo "waiting for couchdb..."
     sleep 1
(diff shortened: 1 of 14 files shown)


couchdb/manage
```diff
@@ -216,18 +223,27 @@ start()
 # Stop the service.
 stop()
 {
-  couchdb -d -p $STATEDIR/couchdb.pid
+  echo "Stopping CouchDB service..."
+  for couch_pid in $(ps aux | grep couch | grep -v grep | awk '{print $2}'); do
+    echo "  killing CouchDB process... ${pid}"
+    kill -9 $couch_pid
```
amaltaro, line 229: For my own reference, this logic to kill the couchdb process in my private VM does not work. It fails and it exits the script, thus never starting CouchDB. Here is the relevant terminal output:
```
+ sudo -H -u $u bashs -lc '/data/srv/current/config/couchdb/manage stop '\''I did read documentation'\'''
Stopping CouchDB service...
  killing CouchDB process... 
/data/srv/current/config/couchdb/manage: line 229: kill: (11496) - Operation not permitted
  killing CouchDB process... 
/data/cfg/admin/InstallDev: line 89: 11496 Killed                  sudo -H -u $u bashs -lc "$cmd"
```

wmagentpy3/local.ini
```diff
@@ -1,61 +1,45 @@
 ; CouchDB Configuration Settings
 
-; Custom settings should be made in this file. They will override settings
-; in default.ini, but unlike changes made to default.ini, this file won't be
-; overwritten on server upgrade.
-
 [couchdb]
 ;max_document_size = 4294967296 ; bytes
+uuid = 652fb71bc6c32121e0752325a4973eb3
```
amaltaro, line 5: TODO: check if it's really necessary.

wmagentpy3/manage
```diff
@@ -492,10 +483,12 @@ start_couch(){
...
-    couchdb -b -a $CONFIG_COUCH/local.ini \
-            -p $INSTALL_COUCH/logs/couchdb.pid \
-            -e $INSTALL_COUCH/logs/stderr.log \
-            -o /dev/null
+    echo -n "Which couchdb: "
+    which couchdb
+    echo "  With installation directory: $INSTALL_COUCH"
+    echo "  With configuration directory: $CONFIG_COUCH"
+    nohup couchdb -couch_ini $CONFIG_COUCH/ \
+        &> $INSTALL_COUCH/logs/couch_nohup.log &
```
amaltaro, line 491: TODO: it's likely a good idea to rename this `couch_nohup.log` log by the usual `couch.log`

couchdb/manage
```diff
@@ -36,11 +36,15 @@ ROOT=$(cd $(dirname $0)/../.. && pwd)
 CFGDIR=$(cd $(dirname $0) && pwd)
 LOGDIR=$TOP/logs/$ME
 STATEDIR=$TOP/state/$ME
+USERPASS=admin:myadminpasswd
```
vkuznet, line 39: You should not define your actual values here since manage script is pushed to public github repo where everyone can find it. Instead you need to check that some env variable exists, e.g.
```
if [ -z "$COUCH_CREDENTIALS" ]; then
   echo "please define $COUCH_CREDENTIALS"
   exit 1
fi
```
This block will check existance of `COUCH_CREDENTIALS` (the name does not matter) which will be setup either by deployment script (post action) or in k8s via secrets. This will allow to keep credentials out of this manage script

couchdb/manage
```diff
@@ -216,18 +223,28 @@ start()
...
 status()
 {
-  couchdb -p $STATEDIR/couchdb.pid -s
-
-  curl -s localhost:5984 | grep -q '"version":"1.[0-9.]*"' ||
+  curl -s localhost:5984/_up | grep -q '"status":"ok"' ||
     { echo "CouchDB is not correctly responding to requests"; return; }
 
-  local TASKS=$(curl -s localhost:5984/_active_tasks)
+  local TASKS=$(curl -s $USERPASS@localhost:5984/_active_tasks)
```
vkuznet, line 247: if you fix this then use URI schema, i.e.
`curl -s http://$USERPASS@localhost:5984/_active_tasks`

couchdb/manage
```diff
@@ -242,7 +259,7 @@ clean_views()
   [ -n "$database" ] ||
     { echo "You must specify the database you wish to clean the views "; exit 1; }
 
-  curl -s -H "Content-Type: application/json" -X POST http://localhost:5984/$database/_view_cleanup | \
+  curl -s -H "Content-Type: application/json" -X POST http://${USERPASS}@localhost:5984/$database/_view_cleanup | \
```
vkuznet, line 262: if you adapt `couchCurl` definition, then this call will be replaced as
```
curl -s -H "Content-Type: application/json" -X POST $couchHost/$database/_view_cleanup | \
```

couchdb/monitoring.ini
```diff
@@ -7,7 +7,7 @@ LOG_ERROR_REGEX='CRASH REPORT'
 PS_REGEX="/bin/beam.smp .*/config/couchdb//local.ini.* -heart"
 
 # The ping test fetches the provided URL and look for the following perl regex
-PING_URL="http://localhost:5984/workqueue/_design/WorkQueue/_rewrite/"
+PING_URL="http://admin:myadminpasswd@localhost:5984/workqueue/_design/WorkQueue/_rewrite/"
```
vkuznet, line 10: This is bad from security point of view as this file will be pushed to github. Instead, as I proposed for manage script you should rely on env variable and read from it couch credentials.

frontend/cmsauth.pm
```diff
@@ -805,6 +805,11 @@ sub authz_complete($$$%)
 		       (map { (length($_), length($hdrs{$_})) } @h),
 		       (map { ($_, $hdrs{$_}) } @h));
     $hdrs{"cms-authn-hmac"} = secret_hmac("authz-headers", $auth);
+    # Alan: make sure these headers are sent to the backend to allow
+    # CouchDB 3.x database access via Proxy Authentication
+    $hdrs{"x-auth-couchdb-username"} = "admin";
```
vkuznet, line 810: This is security flaw again, I suggest to read values from env variables, instead of hard-code them

vkuznet (changes requested):
Alan, I made few suggestions:
- there is a pattern of using `http://$USERPASS@localhost:5984`, therefore it is better to define it once and use across the code. For instance, if it was already there then you don't need to change so many places.
- you should replace all hard-coded couch credentials in all files since you'll push these files into github and we should avoid leakage of credentials (even for our backends)
- usage of perl is not required but leads to increase size of RPMs, therefore I suggest to replace it with sed.

vkuznet:
Regarding security measurements, according to Couch docs we can define group members via simple PUT HTTP request:
```
curl http://localhost:5984/OUR_DB/_security \
-X PUT -H 'content-type: application/json' \
-H 'accept: application/json' \
-d '{"admins":{"names":["dmwm"],"roles":["admins"]},"members":{"names": ["user1","user2"],"roles": ["developers"]}}'
```
Here, we create new users for `OUR_DB` database. The `dmwm` user has role `admins`, while `user1,user2` have role `developers`. I suggest to make separate shell script with this curl template and add it to repository. Then, during deployment phase we can use a script with provided parameters to add new members, e.g.
```
#!/usr/bin/env bash
db=$1
name=$2
role=$3
# perform update in couchdb to assign new member to a given database
curl http://localhost:5984/$db/_security \
   -X PUT -H 'content-type: application/json' \
   -H 'accept: application/json' \
   -d '{"admins":{"names":["$name"],"roles":["$role"]}}`
```

workqueue/monitoring.ini
```diff
@@ -7,7 +7,7 @@ LOG_FILES='*.log'
 LOG_ERROR_REGEX='(Traceback|Exception|Backtrace|CRITICAL|PostMortem|ERROR)'
 
 # The ping test fetches the provided URL and look for the following perl regex
-PING_URL="http://localhost:5984/workqueue/_design/WorkQueue/_rewrite/_show/status/task_activity"
+PING_URL="http://admin:myadminpasswd@localhost:5984/workqueue/_design/WorkQueue/_rewrite/_show/status/task_activity"
```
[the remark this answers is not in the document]
amaltaro: This isn't a bash file, I guess loading credentials from a file name won't work. But I will give it a try.

wmagentpy3/manage
```diff
@@ -444,20 +444,22 @@ db_prompt(){
...
+    echo "  With installation directory: $INSTALL_COUCH"
+    echo "  With configuration directory: $CONFIG_COUCH"
     mkdir -p $INSTALL_COUCH/logs
     mkdir -p $INSTALL_COUCH/database
     perl -p -i -e "s{deploy_project_root/couchdb}{$INSTALL_COUCH}" $CONFIG_COUCH/local.ini
     # couch ini file requires IP based hostname
-    perl -p -i -e "s{bind_address = 0.0.0.0}{bind_address = $COUCH_HOST}g" $CONFIG_COUCH/local.ini
-    perl -p -i -e "s{port = 5984}{port = $COUCH_PORT}g" $CONFIG_COUCH/local.ini
-    perl -p -i -e "s{;admin = mysecretpassword}{$COUCH_USER = $COUCH_PASS}g" $CONFIG_COUCH/local.ini
+    perl -p -i -e "s{bind_address = 127.0.0.1}{bind_address = $COUCH_HOST}g" $CONFIG_COUCH/local.ini
```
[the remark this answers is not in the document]
amaltaro: About replacement of perl by sed, I'd rather leave it for the next step, once everything is functional, tested and deployed. I am being as conservative as I can with these changes.

frontend/cmsauth.pm
vkuznet: this a perl script, therefore it can reads env variables (even though I don't know how it is done in perl). Therefore, I suggest to replace `"admin"` with perl function which will read `COUCH_ADMIN` environment variable, if it is not set it should exit with error.

workqueue/monitoring.ini
```diff
@@ -7,7 +7,7 @@ LOG_FILES='*.log'
 LOG_ERROR_REGEX='(Traceback|Exception|Backtrace|CRITICAL|PostMortem|ERROR)'
 
 # The ping test fetches the provided URL and look for the following perl regex
-PING_URL="http://localhost:5984/workqueue/_design/WorkQueue/_rewrite/_show/status/task_activity"
+PING_URL="http://_admin:_admin@localhost:5984/workqueue/_design/WorkQueue/_rewrite/_show/status/task_activity"
```
amaltaro, line 10: TODO: this should be removed before merging. Actually, this ping test is no longer meaningful, since workqueue runs in k8s where couchdb is not available in localhost.

frontend/cmsauth.pm
```diff
@@ -805,6 +805,11 @@ sub authz_complete($$$%)
 		       (map { (length($_), length($hdrs{$_})) } @h),
 		       (map { ($_, $hdrs{$_}) } @h));
     $hdrs{"cms-authn-hmac"} = secret_hmac("authz-headers", $auth);
+    # Alan: make sure these headers are sent to the backend to allow
+    # CouchDB 3.x database access via Proxy Authentication
+    $hdrs{"x-auth-couchdb-username"} = "_admin";
```
amaltaro, line 810: TODO: It looks like Valentin's comment is gone with a change I made in this line. Bottom line is, we need to investigate how to load this value from a file/environment variable.

amaltaro:
@muhammadimranfarooqi Imran, as discussed today, I separated the frontend changes (and a TODO monitoring.ini) in this PR: https://github.com/dmwm/deployment/pull/1151

Could you please review and merge this one?

couchdb/monitoring.ini
```diff
@@ -7,7 +7,7 @@ LOG_ERROR_REGEX='CRASH REPORT'
 PS_REGEX="/bin/beam.smp .*/config/couchdb//local.ini.* -heart"
 
 # The ping test fetches the provided URL and look for the following perl regex
-PING_URL="http://localhost:5984/workqueue/_design/WorkQueue/_rewrite/"
+PING_URL="http://@@@USER:@@@PASS@localhost:5984/workqueue/_design/WorkQueue/_rewrite/"
```
amaltaro, line 10: rollback

amaltaro:
All right! I think I resolved all the questions and concerns in this PR, other than the migration from perl to sed that goes beyond the scope of these changes.

@muhammadimranfarooqi I have all these things tested in my VM and from my side they are ready to go. Note that it does not require any changes to the frontends!

Regarding testbed deployment, please do not proceed with it because we will have to run the first upgrade together. Actions need to be taken to migrate the databases to a CouchDB 3.x format. It would be great to have it merged sooner though.

PR #1093, opened 2021-09-27
Run CouchDB compaction_daemon for T0 WMAgent

amaltaro:
This PR enables the CouchDB `compaction_daemon` for T0 WMAgent, and also provides some reasonable default thresholds for database and view compaction. This is aligned with the same setup we have for WMAgents and it's been working greatly.

@germanfgv German, could you please change the T0 CouchDB configuration, stop components + couch, start couch + components? At 20:00, it should trigger an automatic compaction of the tier0_request_local db.

### DIFF
--- a/tier0/local.ini
+++ b/tier0/local.ini
@@ -57,3 +57,12 @@ http_connections = 10
 worker_batch_size = 2000
 max_replication_retry_count = infinity
 socket_options = [{keepalive, true}, {nodelay, true}]
+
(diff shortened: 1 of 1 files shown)


germanfgv:
I updated the production config file and compaction is working as expected.

germanfgv:
Hi @muhammadimranfarooqi 
Could you please merge this PR into master?

PR #1109, opened 2021-11-23
Enable GEM DCS bits in DQM GUI plot

pmandrik:
Labels in the vector used for filtering:
const std::vector<std::string> DQMInfoRenderPlugin::summaryMapOrder_
need to be in line with labels in the definition of hist:
https://github.com/cms-sw/cmssw/blob/db63563e447104cef63ed4661c230c4ad1049930/DQMServices/Components/plugins/DQMProvInfo.cc#L177-L178

FYI @quark2


muhammadimranfarooqi:
Hi @pmandrik 


It has been deployed in testbed. Validation may be provided at https://gitlab.cern.ch/cms-http-group/doc/-/issues/282. Thanks

PR #1114, opened 2021-12-07
Adjust FE rules

muhammadimranfarooqi:


frontend/app_dbs_nossl.conf
```diff
@@ -2,5 +2,9 @@ RewriteRule ^(/dbsproxy(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUER
 RewriteRule ^(/dbs2go(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUERY} [R=301,NE,L]
 RewriteRule ^(/dbs2go-writer(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUERY} [R=301,NE,L]
 RewriteRule ^(/dbs2go-migrate(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUERY} [R=301,NE,L]
+
+# this rule is for Python DBS server
```
vkuznet, line 6: put here comment like:
```
# this rule is for accessing DBS server (either Python or GO based)
```

frontend/app_dbs_nossl.conf
```diff
@@ -2,5 +2,9 @@ RewriteRule ^(/dbsproxy(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUER
...
 RewriteRule ^(/dbs(/.*)?)$ https://%{SERVER_NAME}${escape:$1}%{env:CMS_QUERY} [R=301,NE,L]
+
+# This rule is for GO DBS server
```
vkuznet, line 9: and here put comment like:
```
# This rule is only for testing Go DBS server, it should be accompany with appropriate rules from app_dbs_ssl.conf
```

vkuznet: and you'll need to comment out the actual rule since we no longer need /dbsnew end-point

vkuznet (changes requested):

frontend/app_dbs_nossl.conf
muhammadimranfarooqi: `/dbsnew` end-point is still used for DBSWriter

PR #1117, opened 2021-12-16
Stop using passwords for mariadb users

goughes:
Fixes https://github.com/dmwm/WMCore/issues/10926

MariaDB 10.4 changed the default authentication method to allow password-less connections via the local socket file. Details on the change can be found at https://mariadb.com/kb/en/authentication-from-mariadb-104/. 

This PR removes usage of MYSQL_PASS to authenticate to the database, as it is no longer necessary. It also removes the creation of the password for the root user.

More details here: https://github.com/dmwm/WMCore/issues/10837#issuecomment-996082271

### DIFF
--- a/wmagent/manage
+++ b/wmagent/manage
@@ -36,7 +36,6 @@ ORACLE_PASS=
 ORACLE_TNS=
 
 MYSQL_USER=
-MYSQL_PASS=
 if [ -z "$DBSOCK" ]; then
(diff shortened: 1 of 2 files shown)


amaltaro:
I tested these changes with a Py3 MariaDB agent (1.5.7.pre2) and everything seems to be fine!
@muhammadimranfarooqi could you please get this merged?

PR #1119, opened 2021-12-21
Allow WMAgent MariaDB user creation when initializing the database

amaltaro:
This reverts some of the changes made in this PR: https://github.com/dmwm/deployment/pull/1117

Reason is, for WMAgent unit tests setup (docker), we still have to create a third user (different than root and current unix user, at the time agent was deployed).

PR #1120, opened 2021-12-21
Run mysql commands with local current unix user

amaltaro:
This fixes the use of root account to perform MariaDB commands (it is only allowed if we are logged in as root or run those commands as sudo). In addition to that, when installing the database, use the current unix user instead of what is provided in the environment file.

Complement to:
https://github.com/dmwm/deployment/pull/1117
and
https://github.com/dmwm/deployment/pull/1119

@goughes could you please have a look at it as well? This is going together with: https://github.com/dmwm/WMCore/pull/10934

### DIFF
--- a/wmagent/manage
+++ b/wmagent/manage
@@ -170,7 +170,7 @@ load_secrets_file(){
 
     # database settings (mysql or oracle)
     if [ "x$MATCH_ORACLE_USER" == "x" ]; then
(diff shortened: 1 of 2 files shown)


amaltaro:
This has been tested and we can now successfully build wmagent docker images. It's good to go.

@muhammadimranfarooqi if you could merge it still tonight, it would be extremely helpful, since our CI jenkins setup is currently broken. Thanks!

PR #1126, opened 2022-02-16
remove crontab

novicecpp:
Remove cronjob that make CRABServer REST restart every 4 days.
Now, we are deploying CRABServer REST on k8s cluster, we can rely on its memory limit mechanism to kill app when limit are reached.

### DIFF
--- a/crabserver/deploy
+++ b/crabserver/deploy
@@ -39,14 +39,9 @@ deploy_crabserver_sw()
 deploy_crabserver_post()
 {
   case $host in vocms013[89] | vocms073[89] | vocms0143 | vocms074[0123456] | vocms076[12345] | vocms0307 | vocms0318 ) disable ;; * ) enable ;; esac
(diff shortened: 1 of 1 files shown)


novicecpp:
@belforte @mapellidario please review.

belforte:
@novicecpp did you manage to test this change ? I simply do not understand these lines and
can not say it they will be a probelm or not
```
  (mkcrontab
   sysboot

  ) | crontab -
}
```

novicecpp:
I do not understand it too, but I saw it in this committed https://github.com/dmwm/deployment/commit/af6e9a8442b08bab457b67fdb7e9238cc43920a6. It only add restart cronjob after sysboot. So I think it safe for just remove those out. 

Also, I managed to test it and it works.

belforte:
Sounds fine. @muhammadimranfarooqi can you please merge ?

PR #1129, opened 2022-03-01
Remove rse_type based redirects from the frontend.

todor-ivanov:
Fixes https://github.com/dmwm/WMCore/issues/10942

### DIFF
--- a/frontend/backends-k8s-preprod.txt
+++ b/frontend/backends-k8s-preprod.txt
@@ -44,6 +44,9 @@
 ^/auth/complete/ms-monitor(?:/|$) ms-monitor.dmwm.svc.cluster.local
 ^/auth/complete/ms-output(?:/|$) ms-output.dmwm.svc.cluster.local
 ^/auth/complete/ms-rulecleaner(?:/|$) ms-rulecleaner.dmwm.svc.cluster.local
+^/auth/complete/ms-unmerged/(?:.*rse_type=t1)(?:/|$) ms-unmer-t1.dmwm.svc.cluster.local
+^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3us)(?:/|$) ms-unmer-t2t3us.dmwm.svc.cluster.local
+^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3)(?:/|$) ms-unmer-t2t3.dmwm.svc.cluster.local
 ^/auth/complete/ms-unmerged(?:/|$) ms-unmer-t1.dmwm.svc.cluster.local
 ^/auth/complete/t0wmadatasvc(?:/|$) t0wmadatasvc.tzero.svc.cluster.local
 ^/auth/complete/scheddmon/068/ vocms068.cern.ch
--- a/frontend/backends-k8s-preprod.txt
+++ b/frontend/backends-k8s-preprod.txt
@@ -44,9 +44,6 @@
 ^/auth/complete/ms-monitor(?:/|$) ms-monitor.dmwm.svc.cluster.local
 ^/auth/complete/ms-output(?:/|$) ms-output.dmwm.svc.cluster.local
 ^/auth/complete/ms-rulecleaner(?:/|$) ms-rulecleaner.dmwm.svc.cluster.local
-^/auth/complete/ms-unmerged/(?:.*rse_type=t1)(?:/|$) ms-unmer-t1.dmwm.svc.cluster.local
-^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3us)(?:/|$) ms-unmer-t2t3us.dmwm.svc.cluster.local
-^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3)(?:/|$) ms-unmer-t2t3.dmwm.svc.cluster.local
 ^/auth/complete/ms-unmerged(?:/|$) ms-unmer-t1.dmwm.svc.cluster.local
 ^/auth/complete/t0wmadatasvc(?:/|$) t0wmadatasvc.tzero.svc.cluster.local
 ^/auth/complete/scheddmon/068/ vocms068.cern.ch
--- a/frontend/backends-k8s-prod.txt
+++ b/frontend/backends-k8s-prod.txt
@@ -32,9 +32,6 @@
 ^/auth/complete/ms-monitor(?:/|$) cmsweb-srv.cern.ch
 ^/auth/complete/ms-output(?:/|$) cmsweb-srv.cern.ch
 ^/auth/complete/ms-rulecleaner(?:/|$) cmsweb-srv.cern.ch
-^/auth/complete/ms-unmerged/(?:.*rse_type=t1)(?:/|$) cmsweb-srv.cern.ch
-^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3us)(?:/|$) cmsweb-srv.cern.ch
-^/auth/complete/ms-unmerged/(?:.*rse_type=t2t3)(?:/|$) cmsweb-srv.cern.ch
 ^/auth/complete/ms-unmerged(?:/|$) cmsweb-srv.cern.ch
 ^/auth/complete/scheddmon/068/ vocms068.cern.ch
 ^/auth/complete/scheddmon/069/ vocms069.cern.ch


todor-ivanov:
Hi  @amaltaro, Could you please take a look at this PR before @muhammadimranfarooqi merge it.

amaltaro:
@todor-ivanov thanks for providing this patch. It needs to be corrected though, and I think there is one missing piece that we need to dig from the commit history and/or discussions with Imran (a subtle specific behavior to the production cluster).
I will look into that and possibly come up with a new PR in the coming hours.

frontend/backends-k8s-preprod.txt
```diff
@@ -44,9 +44,6 @@
 ^/auth/complete/ms-monitor(?:/|$) ms-monitor.dmwm.svc.cluster.local
 ^/auth/complete/ms-output(?:/|$) ms-output.dmwm.svc.cluster.local
 ^/auth/complete/ms-rulecleaner(?:/|$) ms-rulecleaner.dmwm.svc.cluster.local
-^/auth/complete/ms-unmerged/(?:.*rse_type=t1)(?:/|$) ms-unmer-t1.dmwm.svc.cluster.local
```
amaltaro, line 47: @muhammadimranfarooqi Imran, would we have an alias to reach all (or any) of the ms-unmerged pods? What I am trying to see is whether we could have a backend rules similar to those in production but without this `rse_type=blah` piece.

amaltaro (changes requested):
@todor-ivanov actually, since you have already these changes in place, we better follow them up over here.

I think we also have to update the CMSKubernetes repo, this change seems to be no longer needed to me:
https://github.com/dmwm/CMSKubernetes/commit/cb5af9319bd43b123c5a4efffa956eeabd4162d6#

or at least not the part with the query string `?rses=blah`, I guess we don't even support that query anymore in the service(?)

todor-ivanov:
>  I guess we don't even support that query anymore in the service(?)

I think you are correct. And the `status` API is actually behaving not as expected. It always forwards to the T1 service instance regardless of what you put in the query string [1]. Which to me sounds like it always matches only the last option from the list here: https://github.com/dmwm/CMSKubernetes/commit/cb5af9319bd43b123c5a4efffa956eeabd4162d6#diff-899fc0f7277c3e548a3a3d8be05c275768502f805e2d596e1a38355f8aa18d88R56-R59

@muhammadimranfarooqi could you please help us remove all the extra redirects and let the forntend to chose randomly from any service instance who is going to serve both APIs:
`ms-unmerged/data/status?.*` && `ms-unmerged/data/info.*`  

[1]
https://cmsweb.cern.ch/ms-unmerged/data/status?rses=t2t3us
```
{"result": [
 {
  "wmcore_version": "2.0.0.cmsweb1",
  "microservice_version": "2.0.0.cmsweb1",
  "microservice": "MSManager",
  "status": "OK",
  "thread_id": "",
  "start_time": "Thu, 03 Mar 2022 15:24:33 UTC",
  "end_time": "Thu, 03 Mar 2022 15:24:38 UTC",
  "execution_time": 5.067666,
  "error": "",
  "total_num_rses": 7,
  "total_num_files": 0,
  "num_rses_cleaned": 0,
  "num_files_deleted": 0
}]}

```

muhammadimranfarooqi:
@todor-ivanov 

Random redirects from any service instance using same path `ms-unmerged/data/status` && `ms-unmerged/data/info` is not possible as per my understanding. There should be different path for all instances. In simple words, same path cannot be shared across multiple services randomly. I already tested this when these rules were created first time.

amaltaro:
@muhammadimranfarooqi let me try to put it in a different way.

When a end-user requests anything from ReqMgr2 (or DBSServer), it does not go always to the same replica/pod, but it's actually load balanced (though nginx?) among all the instances of that service. If that is corrrect - which it must be - then is there any reason why we cannot do the same for ms-unmerged? If there is a reason, then I would say it lies on the fact that we actually have different "services", e.g.: ms-unmerg-t1, ms-unmerge-t2t3, etc...

The final goal with these changes should be: any request made against `/ms-unmerged` can be served by any of the `ms-unmerged` replicas, ideally balancing the load among all the replicas available in the cluster.

muhammadimranfarooqi:
@amaltaro 

In case of ms-unmerged, there are 3 different services. 

```
[mimran@lxplus8s14]~% kubectl get svc -n dmwm | grep ms-unm
ms-unmer-t1       ClusterIP   10.254.123.213   <none>        8242/TCP,18242/TCP   23d
ms-unmer-t2t3     ClusterIP   10.254.37.213    <none>        8242/TCP,18242/TCP   23d
ms-unmer-t2t3us   ClusterIP   10.254.186.72    <none>        8242/TCP,18242/TCP   23d
```
If you have a single service with multiple replicas, then surely nginx will redirect to the different pods randomly. But in case of different service it cannot.

amaltaro:
I see, that's exactly what I expected.

I guess we will - eventually - have to get back to this logic of which RSEs are done in each replica, merge all those services into a single one, and deal with the service configuration in a different manner.

The `rse_type` for the testbed cluster is still relevant though (@todor-ivanov , please see my comment in the PR and update it accordingly).

@muhammadimranfarooqi Imran, should we remove those outdated lines from CMSKubernetes? If so, will you do so or should Todor make another PR and link it to this discussion?

todor-ivanov:
Hi @amaltaro 
> The rse_type for the testbed cluster is still relevant though (@todor-ivanov , please see my comment in the PR and update it accordingly).

Sorry, I fail to grasp what you mean here. Why would we have the `rse_type` API for `testbed` but not for `production`? And something more, provided it is broken anyway?

todor-ivanov:
Hi @muhammadimranfarooqi  
About: 
> In case of ms-unmerged, there are 3 different services.

Can't we simply not tie the paths to a specific service but instead we try to benefit from resource groups defined in the same namespace, as explained here: https://kubernetes.io/docs/concepts/services-networking/ingress/#resource-backend ?

amaltaro:
For testbed, the map of query string to a given replica is fully functional (because testbed architecture has only one nginx entry). With that, IMO we should keep that logic as a way to reach specific MSUnmerged replicas.

For production, we found out that that is impossible due to the double layer of nginx. So, even though it was a valid change, it never really worked.

todor-ivanov:
@amaltaro  - ok got it. Reverting the change for the  preprod file  now.

@muhammadimranfarooqi - sorry I had to edit my comment above, adding some more info. Please read it again in case you already has.

muhammadimranfarooqi:
Yes, Alan is right. For nginx, query string redirection does not work. Thats why we have only one rule here. 

https://github.com/dmwm/CMSKubernetes/blob/master/kubernetes/cmsweb/ingress/ing-dmwm.yaml#L44-L47

So, are you going to have a single service for all three replicas then?

todor-ivanov:
hi @muhammadimranfarooqi 

> So, are you going to have a single service for all three replicas then?

No we are not yet planning for changing the static scalability configuration and hence moving back to a single service. That's why I asked above if we  could define all the `ms-unmerged-*` services under a resource group under the `dmwm` namespace and then refer to this resource group in the path configuration instead to a single service. Look at my comment above:  https://github.com/dmwm/deployment/pull/1129#issuecomment-1059137240

amaltaro:
Shouldn't we delete the outdated `rses` code mentioned above? From this place:
https://github.com/dmwm/CMSKubernetes/commit/cb5af9319bd43b123c5a4efffa956eeabd4162d6

todor-ivanov:
If we do so, we will make the paths equal. So no way for distinguishing between the different services.  Until we figure out if we can move to resource groups instead of separate services, we need to keep the paths different for each one of them.

It is a different story though, that I think it needs to be `rse_type=.*` instead of `rses=.*`, because we do not forward treat the `rses` this parameter anywhere in the code in contrast to `rse_type`.

amaltaro:
The current changes are looking good to me and it's ready to go.

However, please discuss those ingress rules with @muhammadimranfarooqi tomorrow morning. My basic and possibly wrong understanding is that we only need the last rule out of those 4 rules, thus we would need to keep only:
```
      - backend:
          serviceName: ms-unmer-t1
          servicePort: 8242
        path: /ms-unmerged/data/status 
```

and remove the other 3 ms-unmerged rules.

todor-ivanov:
> My basic and possibly wrong understanding is that we only need the last rule out of those 4 rules,

Yes, that  would do too

todor-ivanov:
And I can see someone has taken care of this 7 months ago so this is already the case: https://github.com/dmwm/CMSKubernetes/blob/9dfe07e2bae1b7705fafd6177cfa981d2e7b0ad0/kubernetes/cmsweb/ingress/ing-dmwm.yaml#L44-L47. Currently all the status/info queries go to the `ms-unmerged-t1` pod. And that also explains my comment here:  [1]


[1]
> And the status API is actually behaving not as expected. It always forwards to the T1 service instance regardless of what you put in the query string

https://github.com/dmwm/deployment/pull/1129#issuecomment-1058232741

todor-ivanov:
@muhammadimranfarooqi  could you please merge this as it is, before cutting and deploying the latest HG2203 tag. Everything what has been discussed here is resolved. We can follow up separately/offline on exploring the options that the resource groups can provide for us. 

Thanks!

PR #1130, opened 2022-03-07
Define MSOutput policy for RelVal; remove legacy tapePledges

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/10106

plus, remove the obsolete `tapePledges` attribute.

### DIFF
--- a/reqmgr2ms/config-output.py
+++ b/reqmgr2ms/config-output.py
@@ -81,23 +81,15 @@ data.sendNotification = SEND_NOTIFICATION
 data.verbose = True
 data.interval = 60 * 60 * 1  # run it every hour
 data.services = ['output']
-# This tape pledges was extracted from WLCG Rebus and will be needed for the PhEDEx-based data placement
-# https://wlcg-rebus.cern.ch/core/pledge/list/?vo_name=CMS&year=2020&type=Tape
-# Note that a few sites are storing HI data as well, so we need to subtract the storage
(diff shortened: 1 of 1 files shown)


haozturk:
@amaltaro  thanks a lot for providing these changes. I have a question: what happens when there are more than 1 destination such as GEN-SIM in this case?

Right now, everything should go to CERN, so I think we should remove FNAL in this PR unless @bbilin has an objection.

bbilin:
https://github.com/dmwm/WMCore/pull/11024

I just wrote exact same suggestion here.

I agree that FNAL shall be droped.

amaltaro:
> @amaltaro thanks a lot for providing these changes. I have a question: what happens when there are more than 1 destination such as GEN-SIM in this case?

@haozturk that's a good question though. I see I could change the WMCore code to make number of copies to be a function of the destinations. So, if someone decides FNAL and CERN - as it is right now - then the Rucio rule should target both destinations but ask for 2 copies instead of only 1. Does this make sense?

Thanks for your input. I am going to remove FNAL from the policy configuration.

haozturk:
> So, if someone decides FNAL and CERN - as it is right now - then the Rucio rule should target both destinations but ask for 2 copies instead of only 1. Does this make sense?

If there is going to be multiple destination feature, I think it's better to ask for a separate copy at every provided location instead of letting Rucio to pick one. It's more explicit. With that, I think we'll not use that feature for the near future, but IMO it's good to have it in case. We should update the documentation about this change. It can be easily confusing.

PR #1132, opened 2022-03-08
Call python3 explicitely in check4db() at manage script for reqmgr2ms

todor-ivanov:


reqmgr2ms/manage
```diff
@@ -79,7 +79,7 @@ fi
 
 check4db ()
 {
-python <<EOF
+python3 <<EOF
```
vkuznet, line 82: @todor-ivanov , may be you can check something like:
```
python -V | grep "Python 3.*"
```
since python may be alias to python3.

vkuznet (changes requested):

reqmgr2ms/manage
todor-ivanov: Thanks @vkuznet. I just added a check for this. Please take another look. Also here are the relative `service_config` changes for the K8 configurations:

https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/135/diffs?commit_id=5fa49174169156d809ff669cd6359fc9d695f5ce

https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/132/diffs?commit_id=ee1505ef2317c6ded954562b10f31791801ddae3

vkuznet:
@todor-ivanov your change does not work on macOS :) it gives me the following:
```
#!/usr/bin/env bash

pythonExec=python
[[ $(python -V 2>&1) =~ Python\ *2.* ]] && pythonExec=python3
#ver=`python -V 2>&1 | awk '{split($0,a,"."); print a[1]}'`
#if [ "$ver" == "Python 3" ]; then
#    pythonExec=python3
#fi
echo $pythonExec
```
and I get
```
/tmp/ttt.sh
python
[~/Work/Languages/Go/gopath/src/github.com/dmwm/dasgoclient, master+14]
vk@vkair(9:53:46)$ python
Python 3.9.10 (main, Jan 15 2022, 12:09:05)
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> ^D
```
If I change the code to
```
cat /tmp/ttt.sh
#!/usr/bin/env bash

pythonExec=python
#[[ $(python -V 2>&1) =~ Python\ *2.* ]] && pythonExec=python3
ver=`python -V 2>&1 | awk '{split($0,a,"."); print a[1]}'`
if [ "$ver" == "Python 3" ]; then
    pythonExec=python3
fi
echo $pythonExec
```
I get correct python
```
vk@vkair(9:55:16)$ /tmp/ttt.sh
python3
```

I suggest that you revisit your changes and try them on all possible platforms, I can test it on macOS if you want.

todor-ivanov:
Hi @vkuznet 

Allow me to disagree, I think my change is working as expected. See, what this line says:
```
[[ $(python -V 2>&1) =~ Python\ *2.* ]] && pythonExec=python3
```

is: 
* if the current default `python` executable points/links to python2, then change the executable name to `python3`

Which is what it is expected to do. While your code block is doing the opposite:
```
if [ "$ver" == "Python 3" ]; then
    pythonExec=python3
fi
```

* if the current default `python` executable points/links to python3, then change the executable name to `python3`

I intentionally avoided using anything but strictly bash internals for this check, so that it should have exactly the same behavior with any bash no matter the OS.

vkuznet (approved):

todor-ivanov:
Thanks @vkuznet

PR #1143, opened 2022-05-25
Added frontend rules for old DQM GUI offline flavours K8s deployment

micsucmed:


muhammadimranfarooqi:
Hi, @micsucmed 

Redirection of all 3 different apps running on different ports to the same backend service `dqmgui.dqm.svc.cluster.local` does not make sense

muhammadimranfarooqi:
@micsucmed 

I suggest to have 3 different apps in this case i.e. `offline-k8`, `relval-k8` and `dev-k8` and deploy them in testbed. Then 3 different backend services rules should be like this:

```
dqmgui-offline-k8.dqm.svc.cluster.local
dqmgui-relval-k8.dqm.svc.cluster.local
dqmgui-dev-k8.dqm.svc.cluster.local
```

micsucmed:
Hi @muhammadimranfarooqi,

Sorry about the incompletion. I have added your suggestions.

PR #1151, opened 2022-06-16
Frontend changes required by CouchDB 3.1

amaltaro:
This PR takes over some of the changes initially provided in https://github.com/dmwm/deployment/pull/1088

NOTE that this is a work in progress!

In short:
* we are defining a new frontend/Apache environment variable that will contain the path to a file with CouchDB user and roles
* the frontend `cmsauth` module will parse that file and load the user and role in memory
* once an HTTP request is successfully authenticated, two extra HTTP headers will be added before that request is passed over to the backends
  *  `x-auth-couchdb-username` will contain the CouchDB user name
  * `x-auth-couchdb-roles` will contain the CouchDB user role
* CouchDB will then use this information to authorize the user to access the databases and many other Couch endpoints.

### DIFF
--- a/frontend/cmsauth.pm
+++ b/frontend/cmsauth.pm
@@ -252,7 +252,7 @@ my %host_exempt = ();
 my $server_name;
 my %vocms;
 my %revoked = (HOST => {}, LOGIN => {}, DN => {});
-my $couch_cred_file = $ENV{'AUTH_COUCHDB'};
+my $couch_cred_file;
 my $couch_username;
 my $couch_userrole;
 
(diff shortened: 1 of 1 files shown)


amaltaro:
This is an almost working feature, however I still see authentication error in the backend. I tried debugging it to see if there was any hidden character together with the credentials, but it doesn't seem to be the case. I also couldn't find a simple way to expand all the headers coming to the backend.

Anyways, I discovered 2 CouchDB parameters that would make its authentication work just like CouchDB 1.6:
admin_only_all_dbs
and
default_security

they have been provided here: https://github.com/dmwm/deployment/pull/1088 and some initial tests were successful.

amaltaro:
DO NOT MERGE IT! I am not closing it yet though because we might decide to do something different in the next month or so.

amaltaro:
@muhammadimranfarooqi Imran, I think this PR can be closed now that we know we don't need these changes. If we ever need it, I can come back to this. Closing it.

PR #1167, opened 2022-07-12
Temporary changes to stop CouchDB requests in production

amaltaro:
Imran, I think these will be the necessary changes to make sure couchdb backends to get any requests from the frontends, but please review. I don't think we need to merge it, it's meant only to better communicate and prepare for the upcoming production upgrade.


amaltaro:
No longer relevant.

PR #1173, opened 2022-07-22
Update stop ps command; set configuration cookie

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/11222
Fixes https://github.com/dmwm/WMCore/issues/11230

CouchDB deployment configuration changes required for version 3.2.2. Changes are:
* update the `ps` command used to stop/kill CouchDB processes
* no longer store a copy of `default.ini` in our deployment repo, instead just use the one that comes with CouchDB package
* define a proper value for `setcookie` CouchDB configuration, in the `vm.args` file, otherwise CouchDB will not start!
  * NOTE that this value needs to be the same in all the nodes of the cluster (not needed in our single instance setup).
* disable `Content-Security-Policy` header for some CouchDB endpoints (otherwise it breaks CouchApps)

UPDATE: for the SetCookie option, please refer to this for further details: https://github.com/apache/couchdb/discussions/4030


vkuznet:
@amaltaro did you consider of using `couchdb stop` instead of `ps`,  at least I can see from stack overflow posting that this should be used, see [here](https://stackoverflow.com/questions/41520478/how-to-properly-start-stop-couchdb-2-0-on-ubuntu) or follow advise from CouchDB documentation to use [runit](https://docs.couchdb.org/en/stable/install/unix.html?highlight=stop%20service#running-as-a-daemon) script.

amaltaro:
Old CouchDB versions used to have some options for the couchdb wrapper script that would make service management easier. In CouchDB 3.x (maybe starting in 2+), many of those options were removed.

Yes, this `runit` seems to provide a simple way to manage the service. But it's another dependency that we need to make available in the couchdb hosts and it seems not to be isolated to the "user" area as well.
As we are considering to move to CouchDB As a Service in the near future, I think it's not worth the investment now.

amaltaro:
I think I am done with these changes. Feel free to review and merge them, @muhammadimranfarooqi

PR #1171, opened 2022-07-21
Changes to the style of HG PCL Alignment plots

mteroerd:
#### Changes:

Changed the plotting style for HG PCL Alignment plots.
1. The 2D status plot  is now plotted in COLZ mode
2. The movement plots have thresholds drawn, and information is displayed if the observed movement would trigger an update and if not, why.
3. Some indentation fixes were (accidentally) performed because there was a mix of tabs and spaces in the file.

The changes are related to this PR in CMSSW: https://github.com/cms-sw/cmssw/pull/38815 

#### Validation:

The changes were found to correctly change the style of the targeted plots. It was also confirmed that this does not affect the plots of the LG PCL alignment.

PR #1183, opened 2022-08-24
Included HG PCL alignment veto logic in DQMGUI plots

mteroerd:
**Changes:**
Follow-up to #1171.

The high granularity PCL alignment has a simple veto logic, which was previously not used but will be in the near future. The plots now indicate if there was a movement that exceeded the maximum movement or error threshold, which would lead to a veto.

**Validation:**

Tested with a running dqm gui and a ROOT file whether the plots look as expected after the changes .

### DIFF
--- a/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
+++ b/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
@@ -400,18 +400,25 @@ private:
 
 
     int alignableCount = 0; // number of alignables
+    // THRESHOLD
     int aboveCutCount = 0; // number of alignables above threshold cut
     int aboveSigCount = 0; // number that fulfill prior condition and are significant
-    int withinMaxMoveCount = 0; // number that fulfill prior condition and within movement bounds
-    int withinMaxErrorCount = 0; // number that fulfill prior condition and within error bounds
-
+    // VETO
+    int aboveMaxMoveCount = 0; // number outside movement bounds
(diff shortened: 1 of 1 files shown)


tvami:
hi @muhammadimranfarooqi it would be very helpful to have merged in asap, until we still have 48hs for the Prompt to finish... after that it's 3 weeks from now where this would be needed

mmusich:
@tvami 

>  it would be very helpful to have merged in asap, until we still have 48hs for the Prompt to finish... after that it's 3 weeks from now where this would be needed

merging this won't really help. The offline GUI is updated monthly with all the other cmsweb services

tvami:
ok I see, well, that's unfortunate then

mteroerd:
Maybe leave this open for a bit, we found a small inconsistency that we need to investigate.

mteroerd:
There was a small mismatch between the nominator for the threshold logic in the HG PCL and here, since for the logic only divides by the number of alignables that were actually aligned. This is now adjusted here.

mteroerd:
Now panels/ladders that were not aligned are shown in the plots and text clarifies how many panels/ladders were aligned. Also, the grey histograms were made darker to be more readable.

muhammadimranfarooqi:
HI @mteroerd 

This PR was deployed in testbed. Could you please confirm is it ready to be deployed in the production server?

mteroerd:
Looks correct!

PR #1188, opened 2022-09-07
Adding new CouchDB VMs to exporters

arooshap:


arooshap:
Hi @amaltaro,

Could you please review this pull request?

Thanks

exporters/manage
```diff
@@ -234,14 +237,29 @@ start()
             reqmgr2ms_monitors
             ;;
 
+        vocms0842 )
+            das2go_monitors
```
amaltaro, line 241: Are we going to run DAS and MongoDB on this node? Do we actually run it in the current node (vocms0742)?
I guess we do not need to monitor reqmgr2ms as well.

exporters/manage
```diff
@@ -220,6 +220,9 @@ start()
         vocms0740 )
             couchdb_monitors "workqueue,workqueue_inbox"
             ;;
+        vocms0841 )
```
amaltaro, line 223: Instead of making a separate case/block for each of these new nodes, how about coding something like:
```
ocms0740 | vocms0841 )
            couchdb_monitors "workqueue,workqueue_inbox"
            ;;
```
?

Same applies for the other new nodes.

amaltaro (changes requested):
@arooshap thank you for providing these changes. Please find a few comments along the code.
The way I understand it, these new VMs will be solely running CouchDB, right? In that case, there are a few exporters that we should not be running on these nodes.

exporters/manage
arooshap: @amaltaro I have made the required changes.

exporters/manage
arooshap: @amaltaro I have made the changes to this one.

amaltaro:
@arooshap I think we still have corrections to be made.
This PR is going to monitor the following databases for all the new nodes:
```
couchdb_monitors "workqueue,workqueue_inbox"
```

but we actually should monitor only a few databases in each new machine.

We currently have this:
```
vocms0740	workqueue, workqueue_inbox
vocms0742	reqmgr_workload_cache, reqmgr_config_cache, acdcserver, reqmgr_auxiliary
vocms0743	wmstats, workloadsummary, wmstats_logdb
vocms0744	tier0_wmstats, t0_workloadsummary, t0_request, t0_logdb
```

for the new nodes, I would suggest the same distribution:
```
vocms0841	workqueue, workqueue_inbox
vocms0842	reqmgr_workload_cache, reqmgr_config_cache, acdcserver, reqmgr_auxiliary
vocms0843	wmstats, workloadsummary, wmstats_logdb
vocms0744	tier0_wmstats, t0_workloadsummary, t0_request, t0_logdb
```

can you please fix it?

PR #1196, opened 2022-10-20
Adding frontend rules for httppy.

arooshap:


muhammadimranfarooqi:
@arooshap 

Please add relevant rules in backends-k8s-preprod and backends-k8s-prod files as well

PR #1197, opened 2022-10-21
TkDQM PCL Alignment: add style for IsVetoed string

dmeuser:
In this PR the style of the IsVetoed string, added in https://github.com/cms-sw/cmssw/pull/39552, is adapted to improvement the readability. The three different options (Update, Veto, N/A) for this string are now shown in three different colors and the font size is increased.

In addition, we fixed the style of the movement plots, which veto an alignment update. The text used in these plots was slightly displaced.

The changes can be checked in a private DQM GUI using ssh -NL 8080:localhost:8080 <username>@lxplus706.cern.ch and following https://tinyurl.com/25pudh39. There are three runs uploaded to the GUI, one for each of the three options mentioned above.

cc: @mmusich, @arossi83

### DIFF
--- a/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
+++ b/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
@@ -69,25 +69,47 @@ public:
   {
     c->cd();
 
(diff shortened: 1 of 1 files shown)


arooshap:
Hi @dmeuser ,

These changes in the PR have been incorporated in the `HG2211b` image in the testbed. You may proceed with the validation. :)

Cheers,
Aroosha

PR #1198, opened 2022-10-25
CouchDB backend rules for the new SSD nodes

amaltaro:
This PR replaces the old VM ceph-based CouchDB backends:
vocms0740, vocms0742, vocms0743 and vocms0744
by new SSD-based VMs named as:
vocms0841, vocms0842, vocms0843, vocms0844.

It also removes a rule for the `wmdatamining` database, which has not been ported to the new CouchDB 3.x version.


muhammadimranfarooqi:
+1

PR #1199, opened 2022-10-25
Remove no longer needed backend map files from the frontend

amaltaro:
I understand that these files were only used when deploying frontends in the bare VM. 
Given that all the frontends have been migrated to the k8s environment, hence using files named like `backends-k8s-*.txt`, I assume these are no longer needed and the best is to remove them to avoid confusion.

@arooshap @muhammadimranfarooqi please let me know if I am doing something wrong here. Otherwise, feel free to merge it.


muhammadimranfarooqi:
@amaltaro 
It looks good. Cheers

PR #1207, opened 2022-11-03
python 2 to 3 for phedex mkauthmap script

Panos512:


phedex/mkauthmap
```diff
@@ -97,11 +97,11 @@ def updateFile(opts, jsondata):
               tmpfile.write(jsondata)
               tmpfile.close()
             except(IOError, OSError) as e: 
-              print "An error ocurred:"
+              print("An error ocurred:")
               pprint(e)
 
             myumask = current_umask()
-            os.chmod(tmpname, 0666 & ~myumask)
+            os.chmod(tmpname, 0o666 & ~myumask)
```
vkuznet, line 104: I think you have type here, plase check `0o666 `

vkuznet (changes requested):

PR #1209, opened 2022-11-04
Adding frontend8443

arooshap:


frontend8443/frontend.conf
```diff
@@ -0,0 +1,156 @@
...
+  @INCLUDE auth.conf@
+  @INCLUDE app_*_ssl.conf@
+</VirtualHost>
+
+# Define server virtual host.
+# Keep this section last to avoid internal dummy connection issues like
+# SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol -- speaking not SSL to HTTPS port!
+# for more information please see this document
+# https://cwiki.apache.org/confluence/display/httpd/InternalDummyConnection#SSL_Considerations
+Listen 80
```
muhammadimranfarooqi, line 129: @arooshap 
I think we don't need this section for port 80 as well. You may remove this as well. This service only listens `8443`

arooshap: Hi,

does it look fine now?

muhammadimranfarooqi: Hi @arooshap 
Yes, it looks fine. Please proceed with the deployment along with other changes I mentioned in the ticket. Cheers

PR #1210, opened 2022-11-04
Removing extra phedex VMs from production

arooshap:


muhammadimranfarooqi:
@arooshap 
It looks good. You may merge it and proceed with the new tag. Cheers

PR #1222, opened 2022-11-30
Add env variable for allowing obsolete OpenSSL.

todor-ivanov:
Fixes https://github.com/dmwm/WMCore/issues/11378  

This PR is to allow obsolete OpenSSL library to be used as backends for `cryptography` v3.2.1. This is supposed to be a temporary solution since starting with v3.3 and later this workaround will no longer be an option.

### DIFF
--- a/reqmgr2ms/manage
+++ b/reqmgr2ms/manage
@@ -66,6 +66,7 @@ COLOR_NORMAL="\\033[0;39m"
 
 . $ROOT/apps/$ME/etc/profile.d/init.sh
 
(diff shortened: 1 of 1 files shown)


amaltaro:
@todor-ivanov we will need the same change for Global Workqueue.
Can you please make another PR updating this file:
https://github.com/dmwm/deployment/blob/master/workqueue/manage
? Thanks

todor-ivanov:
Thanks @amaltaro 
I just did it https://github.com/dmwm/deployment/pull/1223

PR #1228, opened 2022-12-21
Incorporating wflow-dbs in frotnend rules.

arooshap:


vkuznet:
@arooshap , the changes are look fine to me but I think we need additional entry in `backends*.txt` files to specify which backend to use. Feel free to apply it to testbed and try out to reach the service.

arooshap:
@vkuznet Thanks for the heads up. Yes, I will be incorporating the changes in the backends*.txt files, and I will tag you to verify the changes.

PR #1268, opened 2023-05-12
Check for nullpointer in postdraw of HGPCL

mteroerd:
In the current code, no check for nullpointer is made for the strings isVetoed and PedeExitCode. This seems to have sometimes led to ROOT crashes, resulting in "Blacklisted for crashing ROOT" being displayed for these strings.

This PR adds a simple safety check to prevent this from happening.

### DIFF
--- a/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
+++ b/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
@@ -69,46 +69,48 @@ public:
   {
     c->cd();
 
-    if( o.name.find("SiPixelAli")!= std::string::npos){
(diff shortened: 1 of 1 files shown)


arooshap:
Hi @mteroerd , yesterday, I incorporated this PR in the testbed VMs `vocms0132` and `vocms0731` with the latest release HG2305a.

Please validate the changes. :)

mteroerd:
Hi @arooshap It looks fine to me in the testbed, so it does not look like anything was broken by the change (as expected). 
However, we were never able to reproduce the original problem in the first place, so only time will tell if it was actually fixed by this change.

PR #1276, opened 2023-07-21
[Offline DQMGUI] add run number to prioritize file indexing as input argument of visDQ…

syuvivida:
This PR must be deployed together with release [9.8.0 of cms-DQM/dqmgui_prod](https://github.com/cms-DQM/dqmgui_prod/releases). 
In 9.8.0 release, we updated visDQMImportDaemon so that the min run number to give priority of file indexing is an input argument (rather than a hard-wired number). 

Therefore, in manage, for the 2023 runs, one needs to add one more input argument to change the default value of visDQMImpotDaemon when starting the agents of offline DQMGUI (start_agents_offline). This value can be changed in the future every year when we know the first meaningful run numbers.

### DIFF
--- a/dqmgui/manage
+++ b/dqmgui/manage
@@ -422,6 +422,7 @@ start_agents_offline()
       $DQM_DATA/agents/register \
       $DQM_DATA/data \
       $DQM_DATA/ix128 \
+      --min_run 365753 \
       --next $DQM_DATA/agents/qcontrol $DQM_DATA/agents/vcontrol \
       --rsync \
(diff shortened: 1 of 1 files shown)


arooshap:
Hi @syuvivida, your PR has been incorporated in the latest HG2307a version in vocms0731. Kindly validate :)

syuvivida:
Hi @arooshap 
   I have tested the deployment at vocms0731 and also checked out HG2307a at lxplus. In both cases, the priority of file indexing is behaving as what I expected. Thanks!!

Eiko

PR #1283, opened 2023-09-29
Rules for new dqm services.

arooshap:


mrceyhun:
Thanks @arooshap . AFAIU, it seems ok. It would be great if you can deploy Auth proxy settings to testbed and prod, and  @nothingface0 can check if it's working. I'll not have access unfortunately. As far as I can tell, it should work.

PR #1293, opened 2023-11-22
Update WMAgent deployment to parse MSPILEUP_URL - take2

amaltaro:
Deployment updates for WMAgent to go together with https://github.com/dmwm/WMCore/pull/11795

Please do not merge it yet.

@todor-ivanov we need to decide whether:
a) we enforce the new `mspileup_url` parameter, but then it's not backwards compatible.
b) we only pass the `mspileup_url` parameter if it's found in the secrets file, but then we might be deploying agents in the future without seeing that the MSPileup endpoint was not configured in the agent.

I opted for a) for the moment, but if you prefer, I can change it to b). Please let me know.

### DIFF
--- a/wmagentpy3/manage
+++ b/wmagentpy3/manage
@@ -66,6 +66,8 @@ RUCIO_HOST=
 RUCIO_AUTH=
 RUCIO_ACCOUNT=
 
+MSPILEUP_URL=
+
 #
 # Init checks
 #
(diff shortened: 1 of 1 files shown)


todor-ivanov:
hi @amaltaro 

`a)` should be ok I believe. But we'll need an additional issue in https://github.com/dmwm/WMCore/issues/11314 about  adding this same to the Docker Container and the needed configuration as well. Could you please create it and link me there. I can follow on first possibility.

PR #1298, opened 2024-01-22
[DQMGUI] Adapt for new Point 5 machines

nothingface0:
This PR aims to adapt the existing deployment configurations for DQMGUI for the needs of [the updated deployment](https://github.com/cms-DQM/dqmgui_prod_deployment) procedure on RHEL8 with Python 3.8, which should replace the [existing one](https://github.com/dmwm/deployment/blob/master/Deploy) (which does not work for anything newer than CentOS7).

It has currently been tested on Point 5 machines and OpenStack VMs, but not on the CMSWEB `vocms0731`, `vocms0738` or `vocms0739` ones (which are currently running CentOS7 with Python 2), meaning that those changes might **not** be backwards-compatible with the existing [`Deploy`](https://github.com/dmwm/deployment/blob/master/Deploy) script. 

We suppose that the three `vocms073X` machines will be updated to a more modern OS version at some point, meaning that the currently used `Deploy` script will be deprecated, one way or another.

We are opening this PR in order to review and discuss the changes needed in the future.

Dimitris for DQM-DC

### DIFF
--- a/dqmgui/layout-test
+++ b/dqmgui/layout-test
@@ -2,6 +2,7 @@
 from Monitoring.Core.Defs import ServerDef
 import Monitoring.DQM.GUI
 import os, re, sys
+from datetime import datetime
 
 # use: ./layout-test server_config_file MonitorElementsList
 # example:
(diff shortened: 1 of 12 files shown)


nothingface0:
@arooshap Is there any chance for this PR to be tested on `vocms0731`, just to see if those changes can work with the current deployment procedure?

arooshap:
@nothingface0 sure, I can create a deployment on Monday!

arooshap:
@nothingface0 could you please remove this as a draft, so that I can incorporate these changes?

Issue #1254, opened 2023-03-07
Update of deployment and required packages for RHEL8? 


syuvivida:
Hello,
   We are working on the upgrade of DQM machines. The physical machines are upgraded with more powerful ones and the operation system is also moving from centos 7 to RHEL8. Here we have a couple of questions and wonder if the experts could help us.

1) is there a plan to update the codes (for example the script [Deploy](https://github.com/dmwm/deployment/blob/HG2302d/Deploy) to include the new SCRAM_ARCH  el8_amd64_gcc11? Sorry I am still new to the codes here and not sure how much the SCRAM_ARCH matters here.

2) While we were trying to create a temporary DQMGUI and ran the command at our new DQMGUI servers, we find a lot of packages for bootstrap are missing. Most of them could be installed. However, one of them is not supported by RHEL8: **compat-libstdc++-33**

Do you know if there is a way to resolve this? Thanks!!

$PWD/deployment/Deploy -A slc7_amd64_gcc630 -r "comp=comp" -R comp@HG2209d -t MYDEV -s "prep sw post" $PWD dqmgui/bare 

Shin-Shan Eiko Yu for DQM-DC

muhammadimranfarooqi:
Hi @amaltaro @vkuznet 

Could you please comment. Thanks

amaltaro:
@syuvivida Hi Shin-Shan, I fear we do not have any central effort to move the COMP packages to a new ScramArch (be it a new OS or a new GCC version). 
For the past year or two, we have actually started recommending groups to move away from the RPM-based packaging and deployment, towards the a more standard kubernetes deployment based on docker images for the services. Right now there is a mix of models in the CMS COMP system:
a) some systems managed to move completely to docker + kubernetes (no longer relying in RPMs neither in our specific COMP stack - coupled to SLC7 under gcc630)
b) other services moved to docker + kubernetes, but the docker image is still based on the RPM packages (thus, still completely dependent on the COMP stack)
c) other services have NOT been migrated to docker + kubernetes and they rely fully in the COMP stack. Many of these are legacy and/or orphan services.

I know there has been developments on a new DQM Gui and that their deployment is using kubernetes (and if I am not wrong, built together with CMSSW). So maybe the best recommendation I could give you is to discuss this with the DQM team first and see what the options are.
I let others complement and/or correct what I wrote here.

vkuznet:
[...]

PR #1299, opened 2024-01-29
[DQM] Restore configuration for vocms073* machines

nothingface0:
In continuation of https://github.com/dmwm/deployment/pull/1298, this PR restores some DQMGUI-related configuration specific to `vocms` machines, which are still using [`Deploy`](https://github.com/dmwm/deployment/blob/master/Deploy) for DQMGUI deployment on CC7.

Basically, this PR restores the configuration that existed before it was updated in https://github.com/dmwm/deployment/pull/1298 for those machines.

This is part of our (DQM-DC) effort to move to RHEL8 (mentioned in https://github.com/dmwm/deployment/issues/1254).

Dimitris for DQM-DC

### DIFF
--- a/dqmgui/manage
+++ b/dqmgui/manage
@@ -233,12 +233,23 @@ startagent() {
   esac
   local logstem=${1}
   shift
-  (
-    [ ! -z $edition ] && source $ROOT/apps/dqmgui/$edition/etc/profile.d/env.sh
-    set -x
-    date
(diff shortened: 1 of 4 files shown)


arooshap:
Hi @nothingface0, I deployed the changes in the testbed by editing the HG2401b tag. Please have a look and let me know if we are good to go :)

PR #1302, opened 2024-02-18
Modernize DT DQM GUI layouts and add online DT DQM GUI customisation

battibass:
**PR** prepared as discussed over mattermost with DQM L2 convener (@syuvivida).

Includes the following updates to the DQM GUI:

- modifications to `dt-layouts.py` to make it work in the latest GUI version
- an update to `dqmgui/server-conf-online.py` that applies customisations targeting the deployment of a private DT DQM online GUI instance

Both were tested using a personal deployment of the DQM GUI at P5.
As I am not familiar with the rules for PR testing within this repository, I remain at disposal in case further work is needed on the DT community side.

PR #1316, opened 2024-04-27
Add local trigger information (theta view) to DT layouts and fix  a minor issue in the Private DT DQM GUI customisation

battibass:
Following the workflow applied for PR #1302, here is an additional **PR** that:

- adds to `dt-layouts.py` more plots, upon request of the DT P5 experts 
- fix a small bug to `dqmgui/server-conf-online.py` making it consistent with the actual private DT DQM online GUI depoloyed in `dqmfu-c2b03-46-01`

Both were tested using the private deployment of the DQM GUI at P5.
I remain at disposal in case further work is needed on the DT community side.


nothingface0:
@battibass We've deployed on Playback (https://cmsweb.cern.ch/dqm/online-playback), could you check if it's OK?

battibass:
Thank you very much @nothingface0.
I confirm changes are as expected.

arooshap:
The changes have been incorporated in the release `HG2404a`. Let me know if it is okay for production!

Cheers, 
Aroosha

syuvivida:
Hi @arooshap 
   This PR doesn't affect offline DQMGUI so there is no urgency for pushing for production at vocms0738. Thanks!!

PR #1321, opened 2024-06-06
Update mkauthmap to have the same output format as vomsmap file.

arooshap:


Panos512:
Thanks a lot @arooshap 
This looks great 
CMS-CRIC is running in alma8 since yesterday and I have cleaned up the ~5k old users I told you about. 
I think we can now focus on this.

I have a couple questions and I believe we can merge it and test it.

1. There are some DNs that already contain double quotes and then we wrap them in double quotes. I hope this doesn't break anything:
e.g.
```
"/STREET="Place de l'Université,1"/L=Louvain-la-Neuve/S=Brabant wallon/C=BE/O=Université catholique de Louvain/CN=Agni Bethani" cms
```

2. I'm not sure we actually do anything with the sitemap and the roles. In the end what we get is a list of all CRIC DNs, in quotes with the word cms next to them. I'll comment directly in the code with more details

frontend/mkauthmap
```diff
@@ -106,92 +106,79 @@ def buildSiteMap(sites):
...
     validateMapKeys(struct)
-    # now convert all group/roles to lower
     rolesToLower(struct)
-    # Generate stable JSON representation.
-    enc = json.JSONEncoder(sort_keys=True)
-    jsondata = "[\n " + ",\n ".join(enc.encode(entry) for entry in struct) + "\n]\n"
 
-    # compare this against the current one (create it if it does not exist)
+    # Generate output lines in the required format
+    for entry in struct:
```
Panos512, line 139: and in the end we don't look into the ROLES the NAME the ID or the LOGIN and we just dump the DN, followed by `cms`

Which is more than fine. In the end we need a dump of all the DNs that CRIC know about. I just don't get why we query multiple APIs and do check with the roles when we are not using them. Probably it's some outdated logic that used to work but is now deprecated? 

Do we actually use the roles in any place in CMSWeb?

PR #1348, opened 2025-01-17
Adding SiPixelAliHLTHG to HG workflow check in DQM GUI rendering

phnattla:
This PR adds `"SiPixelAliHLTHG"` to the HG check in the DQM GUI rendering plugin.
This is necessary, since the new HLT-based Pixel Alignments have a new naming scheme starting with `SiPixelAliHLT`, as implemented in CMSSW PR https://github.com/cms-sw/cmssw/pull/46888. 
@mmusich

### DIFF
--- a/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
+++ b/dqmgui/style/PCLPixelAlignmentRenderPlugin.cc
@@ -40,7 +40,7 @@ public:
     c->cd();
 
     // Check if HG alignment is used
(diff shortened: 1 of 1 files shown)


mmusich:
@nothingface0 FYI. Not sure if you can pick it up already in the `dev` GUI, or need cmsweb deployment. Thanks in any case.

arooshap:
@nothingface0 here is the tag in case you need it https://github.com/dmwm/deployment/releases/tag/HG2501a :)

PR #1350, opened 2025-04-09
Update to the PPS layouts

AndreaBellora:
In this PR we made a minor cosmetic update to the presentation of TProfiles, and removed some confusing and duplicated histograms from the shifter layout

### DIFF
--- a/dqmgui/layouts/shift_ctpps_T0_layout.py
+++ b/dqmgui/layouts/shift_ctpps_T0_layout.py
@@ -56,6 +56,19 @@ for sector in sectors:
 
   CTPPSCommonLayout(dqmitems, "Pixel ROC hits vs LS "+sector, *rows)
 
+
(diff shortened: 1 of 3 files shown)


AndreaBellora:
> Thank you @AndreaBellora , let me know when it's ready for testing!

I just confirmed with my colleagues that we are good to go 👍

AndreaBellora:
Thank you very much @nothingface0, I checked the online version and everything looks fine to me. Since our offline is identical to the online, there's nothing to worry about that. I also checked it offline with a deployment on OpenStack and it seemed right.
Everything's good on our side 👍

AndreaBellora:
Not very urgent. We'll need this for the first time at the alignment run, which is scheduled for May 2nd at the moment. Anything before that works for us

AndreaBellora:
Thanks to you!

PR #1351, opened 2025-05-05
Add Tracker Phase 2 MC Validation Workspace

ljuckett:
Adds workspace under Monte Carlo for Phase-2 MC Validation.

File Changes:
dqmgui/workspaces-relval.py - Added MC Tk Phase-2 workspace at slot 31 to be under exisitng Tk MC workspace. Incremented other workspace slot numbers to accommodate. 
dqmgui/layouts/p2tkmc_relval-layouts.py - Created Layouts for MC Tk Phase-2.
dqmgui/server-conf-relval.py - Added p2tkmc_relval-layouts.py to LAYOUTS.

Layouts are numbered:
- 0xx: IT Clusters
- 1xx: IT RecHit
- 2xx: IT TrackingRecHit
- 3xx: OT Clusters
- 4xx: OT L1Tracks
- 5xx: OT RecHit

New workspace in GUI:
RelVal New Workspace GUI

Quick Collection:
RelVal P2 QuickCollection

Included folders (NOTE: this is from a ROOT file generated using phase-2 code that is not in an official CMSSW release yet):
RelVal Validation Folders in GUI


ljuckett:
Changes:
Added more layouts for TEDD rings - These are in folders in Layouts/Phase2Tk/TEDD[1/2][+/-]
Added OT TrackingRecHit layouts
Changed a few layout names to be clearer
Number system change:
0XX - IT Clusters
1XX - IT RecHit
2XX - IT TrackingRecHit
3XX - OT Clusters
4XX - OT RecHit
5XX - OT TrackingRecHit
6XX - OT L1Track

tomalin:
This looks very nice, but I do have one significant comment.
The QuickCollection needs to have sufficiently few histograms that someone can understand in a few minutes if the dataset is OK or not. This PR seems to add about 200 histograms, which is really too many to allow this. (In addition, the histograms will be so small, that it will not be possible to really examine them without painful zooming in and sliding around).
The QuickCollections used at Phase 1 don't look terribly well designed to me, but most of them only have about 20 histograms, which seems to me a better target to aim for. I prefer that we start with too few, and add to them if required later, than start with too many. (In addition, recall that Tracker will ultimately also need histograms for hte offline track reco etc., which may also need to go into the same QuickCollection).

ljuckett:
> This looks very nice, but I do have one significant comment. The QuickCollection needs to have sufficiently few histograms that someone can understand in a few minutes if the dataset is OK or not. This PR seems to add about 200 histograms, which is really too many to allow this. (In addition, the histograms will be so small, that it will not be possible to really examine them without painful zooming in and sliding around). 

I agree there's a lot going on in QC but at the moment we don't have histograms being made that can summarise it better. The QC only has basic Delta X & Y for each barrel layer/ IT endcap ring and nothing for TEDDs (they are on the layout page). If something affects only one layer or only clusters, QC has the bare minimum information available to detect it... 

With the ability to overlay the previous release's plots that Alessandro mentioned and the "play" button on the UI that cycles through the layouts in full screen, I think the validation process would only take 5-10 minutes each time. No zooming or scrolling needed.

Perhaps when we make the TEDD summary plots discussed in the mattermost channel we could also do barrel summary, EPix summary, etc. and make those the QC display. If those summaries become To Dos for months, this current QC will still work.

dqmgui/layouts/p2tkmc_relval-layouts.py
```diff
@@ -0,0 +1,2063 @@
...
+                   'description': "Ring 10 Delta X Pixel"},
+                  {'path':"TrackerPhase2OTTrackingRecHitV/EndCap_Side1/TEDD_1/Ring10/Delta_Y_Pixel",
+                   'description': "Ring 10 Delta Y Pixel"}],
+                 [{'path':"TrackerPhase2OTTrackingRecHitV/EndCap_Side1/TEDD_1/Ring10/Delta_X_Strip",
+                   'description': "Ring 10 Delta X Strip"},
+                  {'path':"TrackerPhase2OTTrackingRecHitV/EndCap_Side1/TEDD_1/Ring10/Delta_Y_Strip",
+                   'description': "Ring 10 Delta Y Strip"}]
+                 )
+
+phase2tkmclayout(dqmitems, "TEDD1-/Ring11/514 - OT TrackingRecHit Endcap MINUS TEDD1 Ring11 Strip Delta X and Y",
```
tomalin, line 1581: Is this new code really adding about 2000 lines of python to the repo, with one line for every single histogram we need to add? Is there no clever way to have loops or add folders of histograms?

dqmgui/layouts/p2tkmc_relval-layouts.py
```diff
@@ -0,0 +1,190 @@
...
+                            phase2tkmclayout(dqmitems, layout_name, (layout[0]+layout[1]), (layout[2]+layout[3]))
+                        else:
+                            phase2tkmclayout(dqmitems, layout_name, (layout[0]+layout[1]), (layout[2]+layout[3]), layout[4])
+                        layout_num = layout_num + 1
+                        layout = []
+
+# ----- END LOOP ------
+# L1 TRACK
+phase2tkmclayout(dqmitems, "600 - OT L1Track Eta Efficiency and Resolution",
+                 [{'path':"TrackerPhase2OTL1TrackV/Nominal_L1TF/FinalEfficiency/EtaEfficiency",
```
tomalin, line 112: Could this be shortened by defining a string equal to "TrackerPhase2OTL1TrackV/Nominal_L1TF/FinalEfficiency"?

ljuckett: I have added a loop for the L1T histograms as well

dqmgui/layouts/p2tkmc_relval-layouts.py
tomalin: looks good.

dqmgui/layouts/p2tkmc_relval-layouts.py
tomalin: good

PR #1356, opened 2025-08-11
[DQM] Add a script to restart DQMGUI's webserver if system is low on memory

nothingface0:
This is a very dirty workaround to the problem of DQMGUI's  webserver possibly leaking memory(?), leading to huge amounts of memory usage.

Since the plan for phase 2 is to deprecate the legacy DQMGUI, this script will patch the problem for the time being.

See also: https://github.com/cms-DQM/dqmgui_prod/issues/41


arooshap:
@nothingface0 is this ready to be merged?

nothingface0:
@arooshap Completely missed this comment, yes!

PR #1360, opened 2025-10-02
[DQM] Silence logrotate when called from daily

nothingface0:
Since the `daily` script is called from `cron`, and
`logrotate` already puts its logs to the proper
log file through the `--log` argument, there's no
need to keep its output on stdout, as it
gets caught by `cron` and it keeps sending emails.

### DIFF
--- a/dqmgui/daily
+++ b/dqmgui/daily
@@ -59,4 +59,4 @@ done
 
 ######################################################################
 # Rotate logs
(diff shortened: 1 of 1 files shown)


uhomsuwa:
Hello,

You can now deploy in production. Here is the release link: https://github.com/dmwm/deployment/releases/tag/HG2604a.

